PR #3, opened 2018-06-21
Make ./run bootstrap work cleanly

chrisvire:
* connect with postgres database to correctly DROP and CREATE db
* only attempt to delete the database if it already exists.


NullVoxPopuli:
noice

PR #8, opened 2018-06-21
Authentication with Auth0

NullVoxPopuli:
Auth0 will allow appbuilder to have _every_ identity provider in the future (at least the providers that auth0 supports).

Changes:
 - add auth0/lock to frontend
 - add auth0 to backend
 - be able to login on frontend

Some info:
Welcome email received upon signup:
image
(email is configurable within auth0)

Default login widget:
image

(this will get customized, but we are waiting on designs)

Things to look at in the future, but are outside the scope of this PR:
 - passwordless
 - other identity providers to tie to each auth0 user.


NullVoxPopuli:
Authenticated:
image

Unauthenticated:
image

note that this "login" route is just one of the "starterpack" routes, and isn't actually doing any login.

The test for this was to see if I could use a JWT to authenticate and get a DWKit form back. :-)

source/OptimaJet.DWKit.StarterApplication/Optimajet.DWKit.StarterApplication.csproj
```diff
@@ -1,45 +1,48 @@
-﻿<Project Sdk="Microsoft.NET.Sdk.Web">
```
NullVoxPopuli: upgrades and whitespace changes... cause I guess csproj doesn't use `\n`? idk. now it does :-)

anywho, add ?w=1 to view the diff without whitespace changes.

source/OptimaJet.DWKit.StarterApplication/Startup.cs
```diff
@@ -37,21 +39,36 @@ public void ConfigureServices(IServiceCollection services)
...
-		options.ExpireTimeSpan = TimeSpan.FromDays(365);
-                options.LoginPath = "/Account/Login/";
+               .AddCookie(options => {
+                    options.ExpireTimeSpan = TimeSpan.FromDays(365);
+                    options.LoginPath = "/Account/Login/";
             });
 
             services.AddMvc(options => {
                 options.Filters.Add(typeof(Security.AuthorizationFilter));
-                options.Filters.Add(
```
NullVoxPopuli: this type of cache isn't valid with the upgrade -- which is fine, because we don't want responses cached anyway

source/SIL.AppBuilder.Portal.Frontend/.babelrc
```diff
@@ -1,10 +1,20 @@
 {
     "plugins": [
+        ["transform-inline-environment-variables", {
+            "include": [
```
NullVoxPopuli: environment variable white list. only these will get passed on to the bundled js code

PR #11, opened 2018-06-27
Organization Invitation UI from admin screen + frontend API infrastructure setup

NullVoxPopuli:
This adds a quick super-admin UI for inviting users, as well as implementing the data layer for the frontend.

image

error (expected), because there isn't a backend endpoint to handle this:
image

-----------------------------------------

I've temporarily disabled api access (behind a HAS_API env var)

image

currently, all data interactions will be local. :)


source/SIL.AppBuilder.Portal.Frontend/package.json
```diff
@@ -11,7 +11,7 @@
     "build": "yarn parcel build src/index.html && cp ./public/* dist/",
     "start:no-proxy": "yarn parcel src/index.html --hmr-port $HMR_PORT -p $PORT",
     "start:dev": "yarn ts-node src/devproxyserver.ts",
-    "start:dev:on-host": "API_HOST=localhost:7081 yarn start:dev",
+    "start:dev:on-host": "PORT=9091 API_HOST=http://localhost:7081 yarn start:dev",
```
NullVoxPopuli: this just matches up non-docker running with docker running

source/SIL.AppBuilder.Portal.Frontend/package.json
```diff
@@ -22,16 +22,24 @@
     "tslint-check": "./node_modules/.bin/tslint-config-prettier-check ./tslint.json"
   },
   "dependencies": {
+    "@orbit/coordinator": "^0.15.14",
```
NullVoxPopuli: these are the deps for the data layer

source/SIL.AppBuilder.Portal.Frontend/src/data/schema.ts
```diff
@@ -0,0 +1,51 @@
...
+        isLocked: { type: 'boolean' },
+
+        timezone: { type: 'string' },
+        localization: { type: 'string' },
+        decimalSeparator: { type: 'string' },
+      },
+      relationships: {
+        ownedOrganizations: { type: 'hasMany', model: 'organization', inverse: 'owner' },
+        organizations: { type: 'hasMany', model: 'organization', inverse: 'users' }
+      }
```
chrisvire: Having difficultly implementing this with JsonApiNotNetCore

source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/admin/invite-organization/display.tsx
```diff
@@ -0,0 +1,71 @@
...
+  render() {
+    const { name, ownerEmail } = this.state;
+
+    return (
+      <div>
+        <form className='ui form'>
+
+          <div className='field'>
+            <label>Organization Owner Email</label>
+            <input type='text' value={ownerEmail || ''} onChange={this.ownerEmailChanged} />
```
giancorzo: This validation is unnecessary you are already setting empty values in the state

source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/admin/invite-organization/display.tsx
```diff
@@ -0,0 +1,71 @@
...
+          </div>
+
+          <div className='field'>
+            <label>Organization Name</label>
+            <input type='text' value={name || ''} onChange={this.nameChanged} />
```
giancorzo: Same here - check last comment

source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/admin/invite-organization/display.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+    await this.props.onSubmit(this.state);
+    this.reset();
+  }
+
+  reset = () => {
+    this.setState({ name: '', ownerEmail: '', expiresAt: tomorrow() });
+  }
+
+  ownerEmailChanged = (e: React.ChangeEvent<HTMLInputElement>) => {
```
giancorzo: This can be inline in the render method especially if is only going to be an update in the state

onChange={(e) => this.setState('ownerEmail',e.target.value)}

source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/admin/invite-organization/display.tsx
NullVoxPopuli: if you backspace all your characters the value becomes undefined, which changes the input from a controlled to uncontrolled component.

source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/admin/invite-organization/display.tsx
NullVoxPopuli: I like to keep the template simple. I agree that having methods for every event handler is obnoxious, though.

I think the solution will ultimately be a helper method that a class can have added to it.

I'm thinking about doing something like this:

```
constructor(...) {
  // ...
  setupTemplateHelpers(this); // assigns some helper methods on `this`
}
//...
onChange={this.mut('ownerEmail')}
```


alternatively, I could try a class decorator that assigns the helpers to the prototype and see if there is an easy to to access those.

thoughts?

PR #13, opened 2018-06-28
Feature/sidebar

giancorzo:


source/SIL.AppBuilder.Portal.Frontend/src/ui/components/header/index.tsx
```diff
@@ -2,5 +2,18 @@ import { connect } from 'react-redux';
...
\ No newline at end of file
+import {
+  toggleSidebar
+} from '@store/user-interface';
+
+const mapStateToProps = ({ ui }) => ({
+  isSidebarVisible: ui.isSidebarVisible
+});
+
+const mapDispatchToProps = {
```
NullVoxPopuli: also, I'm _pretty sure_ that this needs to be the function form, because mapDispatchToProps needs to "map dispatch to props" when the functions within don't use dispatch themselves. example:

```ts
const mapDispatchToProps = dispatch => ({
  toggleSidebar: dispatch(toggleSidebar())
});
```

this way of writing mapDispatchToProps is needed because redux needs to know about the `toggleSidebar` object (returned from `toggleSidebar()` ), which contains the action `type` that redux uses to eventually uses to figure out which  reducer to run for that `type`

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/sidebar/display.tsx
```diff
@@ -0,0 +1,25 @@
...
+export interface IProps {
+  isSidebarVisible: boolean,
+  toggleSidebar: any
+}
+
+class Sidebar extends Component<IProps> {
+
+  render() {
+    return (
+      <div className='sidebar' style={{ display: this.props.isSidebarVisible ? 'block' : 'none' }}>
```
NullVoxPopuli: this logic here should be pulled out of the template, like:

```ts
render() {
  const { isSidebarVisible, toggleSidebar } = this.props;

  const sidebarDisplay = isSidebarVisible ? 'block' : 'none';

  return (
     <div className='sidebar' style={{ display: sidebarDisplay }}>

        <button className='close' onClick={toggleSidebar}>
            X (Close)
        </button>
        
       ...
```

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/layout/layout.scss
```diff
@@ -0,0 +1,20 @@
+.sidebar {
+  height: 100%;
```
NullVoxPopuli: depending on your goals with the sidebar, semantic UI's css should already have stuff like this?
(also, I think you'd want `100vh`, cause 100% height can not work sometimes.. I guess still all depending on intent)

source/SIL.AppBuilder.Portal.Frontend/src/redux-store/user-interface/actions/set-active-menu.ts
```diff
@@ -0,0 +1,13 @@
+import { namespace, State } from '../shared';
```
NullVoxPopuli: I don't think you need `set-active-menu`, because the sidebar should use `react-router`'s `NavLink` for the built-in is-active functionality: https://reacttraining.com/react-router/web/api/NavLink/isactive-func

you'll just need to create route stubs for each of the links (empty route files)

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/sidebar/display.tsx
```diff
@@ -0,0 +1,61 @@
...
+          <button
+            data-test-sidebar-close-button 
+            className='close' 
+            onClick={toggleSidebar}>
+            <Icon name='close' size='large' />
+          </button>
+        </div>
+
+        <Menu pointing secondary vertical>
+          <Menu.Item
```
NullVoxPopuli: have you seen `NavLink`? ;)
https://reacttraining.com/react-router/web/api/NavLink

giancorzo: I fixed this, Combined both semantic and NavLink and remove the setActiveMenu

PR #16, opened 2018-07-02
minor styling, auth guard, general things

NullVoxPopuli:
- adds some helpers for adding the auth'd layout styling
 - adds auth guard helpers
 - tests


source/SIL.AppBuilder.Portal.Frontend/src/lib/toast.ts
```diff
@@ -13,11 +13,11 @@ export interface NotifyOptions {
   };
 }
 export function show(msg: string, options: NotifyOptions) {
-  notify.show(msg, options.type, options.timeout);
+  // notify.show(msg, options.type, options.timeout);
```
NullVoxPopuli: commented these out for now, because we eventually are going to be using our own toast notifications, and not this library.

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/header/display.tsx
```diff
@@ -48,23 +56,27 @@ class Header extends React.Component<Props & RouteComponentProps<{}>, State> {
...
+                  <span className='black-text font-lg'>
```
NullVoxPopuli: Changed this because Josh said that we'll just use the first and last initials of a person's name if they don't have a profile picture.

source/SIL.AppBuilder.Portal.Frontend/src/ui/styles/_base/variables.scss
```diff
@@ -0,0 +1,20 @@
+$sizeUnit: rem;
```
NullVoxPopuli: semantic ui has no style utilities... so I copied some of my own from another project.

I think in the future I want to look at https://tailwindcss.com/

source/SIL.AppBuilder.Portal.Frontend/src/lib/toast.ts
chrisvire: Can this wait until you include own toast notification? When there are errors in DWKit, it is useful to see them.

chrisvire: Well, I guess you left the error ones in place.  Still, it is useful for the moment to see messages from DWKit.

NullVoxPopuli: DWKit doesn't use these ?

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/header/display.tsx
```diff
@@ -48,23 +56,27 @@ class Header extends React.Component<Props & RouteComponentProps<{}>, State> {
...
+                    FL
```
giancorzo: We could add a function or helper to generate the name like

```
const initials = name =>
  name ?
    name
      .match(/\b(\w)/g)
      .join('')
      .substring(0, 2) : "";
```

source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/login/index.tsx
```diff
@@ -1,18 +1,28 @@
...
   state = { data: {}, errors: {} };
   render() {
+    const { history } = this.props;
+
     return (
       <div>
         <h2>Login Route </h2>
 
-        <Login />
+        <Login afterLogin={() => history.push('/tasks')}/>
```
giancorzo: I guess for now it's ok to redirect to task but eventually we should redirect to who ever call the login page to remember where you came from

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/header/display.tsx
NullVoxPopuli: eventually. it'll be a part of hooking up to current user data.

source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/login/index.tsx
NullVoxPopuli: agreed, that's not a recorded requirement, but that would be good behavior to have

PR #17, opened 2018-07-02
WIP: OrganizationMembership

chrisvire:
* Add OrganizationMembership Model
* Add Organization.Owner 
* Add {User,Organization}.OrganizationMemberships

Deep Linking is not supported by JsonApiDotNetCore
```
curl http://localhost:7081/api/users/1?include=organizationMemberships.organizations
{
    "errors": [
        {
            "status": "400",
            "title": "Deeply nested relationships are not supported"
        }
    ]
}
```
There is documentation on extending the Repository so I will have to look into it.
@NullVoxPopuli, I need to know what APIs you need.


NullVoxPopuli:
hmm, that's a huge feature of jsonapi that is missing form that library then.
Looks like there is an open issue on it: https://github.com/json-api-dotnet/JsonApiDotNetCore/issues/39

I can get around it with only one level of `include`, but it just means that the UI will be chattier.

Also, as for as APIs the frontend needs, generically:
 - a resource endpoint for all resources
 - the ability to include any relationship (should be able to be done generically, though you may want to have a whitelist of allowed includes on a resources)
 - 401 on unauthenticated
 - 403 on unauthorized
 - 422 on data issues
 - 405 on when a method is not allowed on an action
 - errors should be sent back as jsonapi error objects (like your snippet for the 400)
 - in development, in may be useful to send 5xx errors as json with stack traces
 - the server must always send a response back, regardless of what sort of error could occur within the server

PR #20, opened 2018-07-09
login styles

NullVoxPopuli:
Login Screen:

image


source/SIL.AppBuilder.Portal.Frontend/package.json
```diff
@@ -44,7 +44,7 @@
     "redux-saga": "^0.16.0",
     "reflux": "^6.4.1",
     "sass-loader": "^7.0.3",
-    "semantic-ui-css": "^2.3.2",
+    "semantic-ui-css": "^2.3.3",
```
giancorzo: Check if Headers dropdown works because I had to revert a preview version because of that

source/SIL.AppBuilder.Portal.Frontend/src/lib/auth0.ts
```diff
@@ -77,13 +77,25 @@ export function isLoggedOut(): boolean {
...
+      autoclose: false,
+      autofocus: true,
+      closable: false,
+      allowShowPassword: false,
+      // socialButtonStyle: 'small',
+      languageDictionary: {
+        title: 'Welcome to Scriptura',
+      },
+      theme: {
+        logo: 'https://software.sil.org/wp/wp-content/uploads/2017/01/2014_sil_logo_80w_96h.png',
```
giancorzo: I guess this is temporary but depending on external assets could eventually break the interface

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/login-custom.tsx
```diff
@@ -0,0 +1,124 @@
...
+  afterLogin: () => void;
+}
+
+export interface IState {
+  isPasswordFieldVisible: boolean;
+  username?: string;
+  password?: string;
+}
+
+const EMAIL_REGEX = /(.+)@(.+)\.(.+)/;
```
giancorzo: const emailPattern = /^(([^<>()\[\]\\.,;:\s@"]+(\.[^<>()\[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/;

This pattern is more complex and ISO complaint if I remember correctly

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/login-custom.tsx
```diff
@@ -0,0 +1,124 @@
...
+    this.props.afterLogin();
+  }
+
+
+  render() {
+    const { mut } = this;
+    const { isPasswordFieldVisible, username, password } = this.state;
+
+    return (
+      <div className='flex h-100 flex-grow justify-content-center align-items-center'>
```
giancorzo: We need to be consistent with styling, even though I like it this is not consistent with the other components

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/login-custom.tsx
```diff
@@ -0,0 +1,124 @@
...
+        <div className='bg-white gray-border'>
+          <div className='p-lg p-l-xl p-r-xl flex-column justify-content-center align-items-center'>
+            <h1 className='bold'>Welcome to Scriptura</h1>
+          </div>
+          <hr />
+          <div className='p-lg'>
+            <form className='ui form m-b-lg' onSubmit={this.onFormSubmit}>
+              <div className='field ui input m-b-lg block w-100'>
```
giancorzo: For consistency, we need to use Semantic UI component, and it's better to use the React component of the library when possible

source/SIL.AppBuilder.Portal.Frontend/package.json
NullVoxPopuli: yup, seems fine, can't open both at the same time.

source/SIL.AppBuilder.Portal.Frontend/src/lib/auth0.ts
NullVoxPopuli: yeah, we don't have a domain or s3 bucket to point at yet

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/login-custom.tsx
NullVoxPopuli: yeah, I have strong opinions about email regex. we really don't need something that complicated, and that regex is garbage anayway. We don't need to perfectly validate domains, we just need the required parts of an email -- the before and after the `@` and the before and after the `.`.

Take a look at this: https://davidcel.is/posts/stop-validating-email-addresses-with-regex/
it takes the idea even further by only requiring an `@`.

The point is that we don't need to be over aggressive in telling people that they're wrong, and we have no idea how email addresses will look in the future. For example: a toooooooooooon of websites break on `+tags` in email addresses... but they are fairly common now to track spam / data selling. example: `email+somewebsite@gmail.com`.

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/login-custom.tsx
NullVoxPopuli: what's not consistent?

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/login-custom.tsx
NullVoxPopuli: that's fair. Since this component isn't used, I'll just paste it as a comment for reference, in case we need it later... I was going down this path, and decided to just use the auth0.lock.js stuff.

NullVoxPopuli:
For reference "Just in case"
```tsx
import { withTemplateHelpers, Mut } from 'react-action-decorators';

import { login as loginToAuth0 } from '@lib/auth0';

export interface IProps {
  afterLogin: () => void;
}

export interface IState {
  isPasswordFieldVisible: boolean;
  username?: string;
  password?: string;
}

const EMAIL_REGEX = /(.+)@(.+)\.(.+)/;

// This is an alternative login implementation that may be used later on down the road.
// However: it is recommended to use auth0's built-in lock UI, as it provides
//          social logins 'for free'.
//          See 'src/ui/styles/auth0-overrides.scss' for auth0-lock styles.
@withTemplateHelpers
export default class Lock extends React.Component<IProps, IState> {
  mut: Mut;
  state = { loggedIn: false, isPasswordFieldVisible: false, username: '', password: '' };

  showPassword = () => {
    this.setState({ isPasswordFieldVisible: true });
  }

  isEmailValid = () => {
    const { username } = this.state;

    return EMAIL_REGEX.test(username);
  }

  isFormValid = () => {
    const { password } = this.state;

    return (password && password.length > 0 && this.isEmailValid());
  }

  onFormSubmit = (e) => {
    e.preventDefault();
    const { isPasswordFieldVisible } = this.state;

    if (isPasswordFieldVisible) {
      this.login();
    } else {
      this.showPassword();
    }
  }

  login = async () => {
    const { username, password } = this.state;

    await loginToAuth0(username, password);

    this.props.afterLogin();
  }


  render() {
    const { mut } = this;
    const { isPasswordFieldVisible, username, password } = this.state;

    return (
      <div className='flex h-100 flex-grow justify-content-center align-items-center'>
        <div className='bg-white gray-border'>
          <div className='p-lg p-l-xl p-r-xl flex-column justify-content-center align-items-center'>
            <h1 className='bold'>Welcome to Scriptura</h1>
          </div>
          <hr />
          <div className='p-lg'>
            <form className='ui form m-b-lg' onSubmit={this.onFormSubmit}>
              <div className='field ui input m-b-lg block w-100'>
                <label>Email Address</label>
                <input required
                  type='email'
                  value={username}
                  onChange={mut('username')}
                  className='ui input w-100'
                  placeholder='example@domain.com'
                 />
              </div>

              { isPasswordFieldVisible && (
                <div className='ui input block w-100'>
                  <label>Password</label>
                  <input required autoFocus
                    type='password'
                    value={password}
                    onChange={mut('password')}
                    className='ui input w-100'
                    placeholder='**********'
                  />
                </div>
              )}
            </form>

            { !isPasswordFieldVisible && (
              <button
                disabled={!this.isEmailValid()}
                className='ui button w-100'
                type='submit'
                onClick={this.showPassword}>
                Next
              </button>
            )}
            { isPasswordFieldVisible && (
              <button
                type='submit'
                disabled={!this.isFormValid()}
                className='ui button w-100'
                onClick={this.login}>
                Login
              </button>
            )}
          </div>
        </div>
      </div>
    );
  }
}

```

PR #27, opened 2018-07-16
Improvement how pictures are displayed when upload

giancorzo:
When you upload pictures that are in portrait I adjust it to try to fit it in a square of 128px

potrait
portrait

Same for landscape

landscape
landscape-upload

### DIFF
--- a/source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/profile/picture/index.tsx
+++ b/source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/profile/picture/index.tsx
@@ -48,9 +48,7 @@ class ImageProfile extends React.Component<IProps, IState> {
 
   render() {
 
-    const { imageData, width, height } = this.state;
-
-    console.log(width, height);
+    const { imageData } = this.state;
 
     return (
       <div data-test-picture-profile>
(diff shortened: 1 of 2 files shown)


source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/profile/profile.scss
```diff
@@ -83,11 +83,27 @@
     }
   }
 
-  .profile-image {
+  .thumbnail {
```
NullVoxPopuli: do we need all of this?
have you seen? 
https://github.com/sillsdev/appbuilder-portal/blob/master/source/SIL.AppBuilder.Portal.Frontend/src/ui/styles/utility/image.scss#L1

it's what the settings page uses

source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/profile/picture/index.tsx
```diff
@@ -6,46 +6,65 @@ export interface IProps {
...
         {!imageData ?
           <div className='default-profile-image'>
             <span>JL</span>
           </div> :
-          <img
-            className='profile-image'
-            src={imageData}
-            data-test-picture-uploaded
-          />
+          <div className='thumbnail'>
```
NullVoxPopuli: this could probably be extracted to its own component (the thumbnail and the initials (when there is no image)), and we could throw it in the header as well (if we can adjust the size).
we could then wrap it in a `withCurrentUser` and auto-populate the initials (if they exist (maybe defaut to `FL` for first/last? or `??` to encourage people to set it to something? idk)

PR #32, opened 2018-07-19
Bugfix: Profile menu would not display | Also: Address awkward profile image due to ineditablity

NullVoxPopuli:
image


NullVoxPopuli:
looks like the tests were always passing, because the dom elements are always there, and css classes control whether or not the dropdown is shown or not.

NullVoxPopuli:
(looks like the wrong class is used on the selector in the tests)

PR #33, opened 2018-07-23
Bugfix/org creating should assign current user

NullVoxPopuli:
- org creation should assign current user
 - when logging in, redirect back to the attempted-to-visit page (this only works if the route requires authentication)
 - fix some frontend request stubbing
 - document why auth0 requests can't be stubbed.


source/OptimaJet.DWKit.StarterApplication/Models/Organization.cs
```diff
@@ -11,13 +11,13 @@ public class Organization : Identifiable
...
-        [Attr("websiteUrl")]
+        [Attr("website-url")]
         public string WebsiteUrl { get; set; }
 
-        [Attr("buildEngineUrl")]
+        [Attr("build-engine-url")]
         public string BuildEngineUrl { get; set; }
 
-        [Attr("buildEngineApiAccessToken")]
+        [Attr("build-engine-api-access-token")]
```
chrisvire: see below: should organizationMemberships be changed to organization-memberships like in user?

NullVoxPopuli: yeah, all the jsonapi stuff should be hyphenated

PR #37, opened 2018-07-24
Org Switcher / Conversion of sidebar to offcanvas

NullVoxPopuli:
- Makes sidebar a responsive offcanvas nav
- Adds Org Switcher to the sidebar

image
image
image
image
image

without the search (currently the search will only show up if there are more than 4 organizations (but disabled for the screenshot))
image


NullVoxPopuli:
Note: backend tests are going to be a separate PR, because they're going to a ton of changes.

source/OptimaJet.DWKit.StarterApplication/Controllers/OrganizationInvitesController.cs
```diff
@@ -1,4 +1,3 @@
-﻿using System;
```
chrisvire: I left the default "using System;" in there when I went through and removed unused namespaces.  I figured that it could be common and it might be weird to type a classname you expect to just be there (during SDD) and not be.

NullVoxPopuli:
no idea why tests are failing on travis. :-\

```
Finished in 1.993 secs / 0.085 secs @ 14:29:42 GMT-0400 (EDT)

SUMMARY:
✔ 35 tests completed
ℹ 1 test skipped
```

I get that locally.

PR #31, opened 2018-07-19
Feature/users

giancorzo:

PR #41, opened 2018-07-25
Implement filtering users by Organization header

chrisvire:


source/OptimaJet.DWKit.StarterApplication/Controllers/UsersController.cs
```diff
@@ -7,12 +7,14 @@
 using Microsoft.AspNetCore.Authentication.JwtBearer;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc;
```
NullVoxPopuli: why are these here?

source/OptimaJet.DWKit.StarterApplication/Repositories/UserRepository.cs
```diff
@@ -6,17 +6,32 @@
...
             ) : base(loggerFactory, jsonApiContext, contextResolver)
         {
+            this.OrganizationContext = organizationContext;
+        }
+
+        public override IQueryable<User> Get()
+        {
+            if (OrganizationContext.InvalidOrganization) return base.Get().Take(0);
+            var users = base.Get().Include(u => u.OrganizationMemberships);
+            if (!OrganizationContext.HasOrganization) return users;
```
NullVoxPopuli: this'll return all users in the database, yeah? 
don't we want to scope it to users in all the organizations the current user is in?

source/OptimaJet.DWKit.StarterApplication/Services/HttpOrganizationContext.cs
```diff
@@ -0,0 +1,51 @@
...
+
+        public HttpOrganizationContext 
+        (
+            IHttpContextAccessor httpContextAccessor,
+            OrganizationService organizationService
+        )
+        {
+            this.HttpContext = httpContextAccessor.HttpContext;
+            this.OrganizationService = organizationService;
+            Update();
```
NullVoxPopuli: rather than calling update in the constructor, the methods / properties should be memoized.

(having side-effects in constructors makes testing more difficult)

source/OptimaJet.DWKit.StarterApplication/Controllers/UsersController.cs
chrisvire: May have needed them at one time.  I hoped that Visual Studio would visually show unused namespaces.

source/OptimaJet.DWKit.StarterApplication/Repositories/UserRepository.cs
chrisvire: The current user can be in multiple organizations.  The organization switcher will select one organization.  Will it be able to select multiple organizations?

Where in the UI will be be viewing a list of users anyway?  Just curious.

source/OptimaJet.DWKit.StarterApplication/Services/HttpOrganizationContext.cs
chrisvire: Explain how it should be different?

source/OptimaJet.DWKit.StarterApplication/Controllers/UsersController.cs
chrisvire: All three of those namespaces are needed.

chrisvire: I submitted a commit that remove unnecessary using statements.

source/OptimaJet.DWKit.StarterApplication/Repositories/UserRepository.cs
NullVoxPopuli: the current requirements are 'one org' or 'all of the orgs a user belongs to'.

that'll be in @giancorzo 's PR: https://github.com/sillsdev/appbuilder-portal/pull/31

source/OptimaJet.DWKit.StarterApplication/Services/HttpOrganizationContext.cs
NullVoxPopuli: for example, for the CurrentUser,
image

tihs defines a getter property that is memoized such that subsequent calls don't make db hits again.

the CurrentUser is stashed in the HttpContext, but could probably just be a private property on that class since scoped services last the lifetime of a request.

source/OptimaJet.DWKit.StarterApplication/Repositories/UserRepository.cs
```diff
@@ -6,17 +6,45 @@
...
             IDbContextResolver contextResolver
             ) : base(loggerFactory, jsonApiContext, contextResolver)
         {
+            this.OrganizationContext = organizationContext;
+            this.CurrentUserContext = currentUserContext;
+        }
+
+        public override IQueryable<User> Get()
+        {
+            if (OrganizationContext.InvalidOrganization) return base.Get().Take(0);
```
NullVoxPopuli: semantically, what is an InvalidOrganization? what does this property represent?

anywho, to return an empty collection, I think you can just do 
```cs
 Enumerable.Empty<type of p>().AsQueryable();
```

chrisvire:
This is ready for review again.

source/OptimaJet.DWKit.StarterApplication/Repositories/UserRepository.cs
```diff
@@ -6,17 +6,45 @@
...
+            {
+                // No organization specified, so include all users in the all the organizations that the current user is a member
+                var currentUser = GetByAuth0Id(CurrentUserContext.Auth0Id).Result;
+                var currentUserOrgIds = currentUser.OrganizationMemberships.Select(o => o.OrganizationId);
+                return base.Get()
+                           .Include(u => u.OrganizationMemberships)
+                           .Where(u => u.OrganizationMemberships.Select(o => o.OrganizationId).Intersect(currentUserOrgIds).Any());
+            }
+            else
+            {
```
NullVoxPopuli: because we are returning in each block, we can omit the `else` part and reduce the number of indentations :)

source/OptimaJet.DWKit.StarterApplication/Repositories/UserRepository.cs
chrisvire: We tried that but got an exception (non async IQueryable?  Or something like that).  We put it back in and it worked.  Will change.

source/OptimaJet.DWKit.StarterApplication/Repositories/UserRepository.cs
chrisvire: OK

source/OptimaJet.DWKit.StarterApplication/Repositories/UserRepository.cs
chrisvire: InvalidOrganization is that we parsed a OrgId (as an integer), but it wasn’t found.  So returning empty.  If you want this to be the same as having an empty value (returning users in the same orgs as current user), that would be fine with me.

PR #42, opened 2018-07-26
Add GivenName, FamilyName to User and populate from Auth0 on new user

chrisvire:
* Add GivenName and FamilyName to User object
* When a new user is created, capture the information from the JWT provided by Auth0.
* If the JWT is for auth0 Username-Password-Authentication account type, use the Auth0 Management API to retrieve the GivenName and FamilyName from the user_metadata


source/OptimaJet.DWKit.StarterApplication/Controllers/BaseController.cs
```diff
@@ -39,14 +41,33 @@ public User CurrentUser
...
+                var user = FindOrCreateCurrentUser();
 
                 HttpContext.Items[CURRENT_USER_KEY] = user;
 
                 return user;      
             }
             
         }
+
+        private User FindOrCreateCurrentUser()
```
NullVoxPopuli: I'd move this directly into the CurrentUser Context since it's more of a concern of there than the _base_ controller.

```cs
private User currentUser = null; // or _currentUser, whatever we decided on earlier
public User GetCurrentUser() 
{
     if (!currentUser)
     {
         currentUser = CurrentUserContext.FindOrCreateCurrentUser();
     }

    return currentUser;
}
```

and then in the base controller, you could just do:

```cs
protected User CurrentUser 
{
  get 
  {
    return CurrentUserContext.GetCurrentUser();
  }
}
```

This also would allow the CurrentUserContext to be injected anywhere and allow any class to have memoized access to the current user.

source/OptimaJet.DWKit.StarterApplication/Services/Auth0ManagementApiTokenService.cs
```diff
@@ -0,0 +1,43 @@
...
+        {
+            get
+            {
+                if (token == null) {
+                    var domainUrl = GetVarOrThrow("AUTH0_DOMAIN");
+                    var clientId = GetVarOrDefault("AUTH0_TOKEN_ACCESS_CLIENT_ID", "");
+                    var clientSecret = GetVarOrDefault("AUTH0_TOKEN_ACCESS_CLIENT_SECRET", "");
+                    var client = new RestClient($"{domainUrl}/oauth/token");
+                    var request = new RestRequest(Method.POST);
+                    request.AddHeader("content-type", "application/json");
```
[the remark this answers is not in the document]
NullVoxPopuli: the package you linked to is the AuthenticationAPI, rather than the ManagementAPI, not sure if that is intentional.

but anyway, this reasoning should be noted in the code, and an issue should be opened in Auth0's github and linked here. :)

README.md
```diff
@@ -31,7 +31,13 @@ the tests, and run them individually.
 ### Backend Notes
 
 - All endpoints should be behind an `api/` path
-
+- Access to Auth0 Management API requires configuation in `.env` (which is in `.gitignore`)
+  - Login to http://manage.auth0.com
+  - Navigate to `APIs` -> `Auth0 Management API` -> `Auth0 Management API (Test Application)`
+  - Get `Client ID` and `Client Secret` values
+  - assign to following variables in `.env`
+    - `AAUTH0_TOKEN_ACCESS_CLIENT_ID`
```
NullVoxPopuli: two A's?

Also, after going to the `Auth0 Management API` ya need to click the `Machine to Machine Applications` tab.

also, what scopes are needed in case we need to create a new token?
looks like you just have `read:users`, `read:users_app_metadata`

source/OptimaJet.DWKit.StarterApplication/Optimajet.DWKit.StarterApplication.csproj
```diff
@@ -49,6 +48,8 @@
     <PackageReference Include="DWKit-ProviderForMSSQL" Version="2.1.1" />
     <PackageReference Include="DWKit-ProviderForPostgreSQL" Version="2.1.1" />
     <PackageReference Include="DWKit-Security" Version="2.1.1" />
+    <PackageReference Include="Auth0.ManagementApi" Version="5.8.0" />
+    <PackageReference Include="RestSharp" Version="106.3.1" />
```
NullVoxPopuli: woah! version 106!

source/OptimaJet.DWKit.StarterApplication/Services/Auth0ManagementApiTokenService.cs
```diff
@@ -0,0 +1,43 @@
...
+                    request.AddParameter("application/json", $"{{\"grant_type\":\"client_credentials\",\"client_id\": \"{clientId}\",\"client_secret\": \"{clientSecret}\",\"audience\": \"{domainUrl}/api/v2/\"}}", ParameterType.RequestBody);
+                    IRestResponse response = client.Execute(request);
+                    if (response.IsSuccessful)
+                    {
+                        dynamic json = JObject.Parse(response.Content);
+                        token = json.access_token;
+                    }
+                    else 
+                    {
+                        Log.Error($"Failed to request token from Auth0: Status={response.StatusDescription}, Content={response.Content}");
```
NullVoxPopuli: for whomever is consuming this property, how will they know if token renewal fails?

source/OptimaJet.DWKit.StarterApplication/Services/HttpCurrentUserContext.cs
```diff
@@ -1,20 +1,62 @@
...
+                if (managementApiClient == null) {
+                    var token = TokenService.Token;
+                    var domainUri = new Uri(GetVarOrThrow("AUTH0_DOMAIN"));
+                    managementApiClient = new ManagementApiClient(token, domainUri.Host);
+                }
+                return managementApiClient;
+            }
+        }
+
+        private User Auth0User {
```
NullVoxPopuli: not sure if you want to denote the difference between an Auth0 User and our users here, but it may be good to point out? idk

source/OptimaJet.DWKit.StarterApplication/Services/HttpCurrentUserContext.cs
```diff
@@ -26,5 +68,69 @@ public class HttpCurrentUserContext : ICurrentUserContext
             }
         }
 
+        public string Email {
```
NullVoxPopuli: If the current user is pulled in to this class, then each of these properties can be proxies to the memoized current user which would greatly reduce the amount of memoization in this class (which would achieve simpler implementation, I think)

source/OptimaJet.DWKit.StarterApplication/Services/HttpCurrentUserContext.cs
```diff
@@ -26,5 +68,69 @@ public class HttpCurrentUserContext : ICurrentUserContext
...
+                }
+                return email;
+            }
+        }
+
+        public string GivenName {
+            get {
+                if (givenName == null) {
+                    var auth = AuthType;
+                    if (string.Compare(auth, "auth0", StringComparison.Ordinal) == 0) {
```
NullVoxPopuli: there is `StartsWith`? https://msdn.microsoft.com/en-us/library/baketfxw(v=vs.110).aspx

source/OptimaJet.DWKit.StarterApplication/Services/HttpCurrentUserContext.cs
```diff
@@ -26,5 +68,69 @@ public class HttpCurrentUserContext : ICurrentUserContext
...
+            }
+        }
+
+        public string FamilyName {
+            get {
+                if (familyName == null) {
+                    var auth = AuthType;
+                    if (string.Compare(auth, "auth0", StringComparison.Ordinal) == 0)
+                    {
+                        try
```
NullVoxPopuli: I don't think this is the best use of try/catch here.
functionality wise, it doesn't get us anything additional to not having it (other than adding to the logs, which I think is more info than we need)

Also, what Exception would be thrown?

Catching the root `Exception` class can accidentally catch unexpected errors where you would usually want those to bubble up.

source/OptimaJet.DWKit.StarterApplication/Controllers/BaseController.cs
chrisvire: HttpCurrentUserContext only read the information from the HttpContext (and Auth0 Management API) and doesn't have any dependencies on Model objects or other services.  It would be an inappropriate coupling.

NullVoxPopuli: CurrentUserContext could directly interact with the AppDbContext instead of UserService, I think that would solve the problem :)

I'd do:

CurrentUserContext - depends on:
 - IHttpContextAccessor
 - Auth0Management
 - AppDbContext

UserService (and any other Service) could then inject the CurrentUserContext without circular injection.

At least, that's what I think would provide the best ergonomics, given all the research / messin' with things so far.

Thoughts?

source/OptimaJet.DWKit.StarterApplication/Services/Auth0ManagementApiTokenService.cs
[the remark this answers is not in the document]
chrisvire: This code requests a token to use the Management API.  It looks like they are working on exposing the functionality in Auth0.net to do this request.  Their current documentation for C# to create this request is to use RestSharp (which we did).
https://auth0.com/docs/api/management/v2/tokens#automate-the-process
I can add some comments to the class to explain this.

So there is no need for an issue.  It is work in progress and not released code.

README.md
chrisvire: Will fix the to A's.  Sigh...
Will document the scopes.

source/OptimaJet.DWKit.StarterApplication/Services/Auth0ManagementApiTokenService.cs
chrisvire: I wasn't quite sure how to handle this.  It will silently fail to update the user's GivenName and FamilyName.  We added exception handling in HttpCurrentUserContext and it will be logged.

source/OptimaJet.DWKit.StarterApplication/Services/HttpCurrentUserContext.cs
chrisvire: It is a private property and the using statement is for Auth0 User.  There are other User objects in other namespaces.  Other option is to not use using.

source/OptimaJet.DWKit.StarterApplication/Services/HttpCurrentUserContext.cs
chrisvire: I explained earlier that it would be an inappropriate coupling to bring in model classes into this class.  This is low level access to data in HttpContent (or derived from it via Auth0 Management API).  It was a conscience decision to layer it this way.

source/OptimaJet.DWKit.StarterApplication/Services/HttpCurrentUserContext.cs
chrisvire: I started with `CompareTo` but VS suggested to change it to this.

source/OptimaJet.DWKit.StarterApplication/Services/HttpCurrentUserContext.cs
chrisvire: If we catch it here, then the GivenName and FamilyName are not assign which is not a big deal.  This is just a convenience and not critical (since the user can enter these in their profile).  If the exception goes out of here it could cause the account creation to fail.  I would prefer to swallow exceptions here.

source/OptimaJet.DWKit.StarterApplication/Services/Auth0ManagementApiTokenService.cs
NullVoxPopuli: In my experience silent failure makes tracking things harder.
Logging isn't bad, but something should be done with the logs. 
like, have you used rollbar? or some other exception / error / log tracking service?

source/OptimaJet.DWKit.StarterApplication/Services/HttpCurrentUserContext.cs
NullVoxPopuli: whatever is fine by me

source/OptimaJet.DWKit.StarterApplication/Services/Auth0ManagementApiTokenService.cs
chrisvire: I have used log entries, but wasn't a fan of it.  We should have some service to collect logs.  

This one is so low value, I am not worried about it.

source/OptimaJet.DWKit.StarterApplication/Services/HttpCurrentUserContext.cs
NullVoxPopuli: ok, I just mean that as-is, it's kinda messy. And because this is the `CurrentUserContext`, I thought the data should be primarily backed by a current user, rather than the `HttpContext` (which contains all info about a request). 

The memoization seems to just be a bit much here. 
In practice, these properties should only be read once, because (and correct me if I'm wrong, I could be totally off on intent here):

request comes in:
 - user has a user record
     - CurrentUser should be looked up via the db, and this class's functionality (as it is currently) is then not used, because all the data exists in the db, and is maintained there

 - user does not yet have a user record
     - This class is used to build the User Entity before saving, in which case all of these attributes, Email, Given Name, etc, are only read once, and memoization is unneeded.

source/OptimaJet.DWKit.StarterApplication/Services/HttpCurrentUserContext.cs
NullVoxPopuli: I mean, I think you can just do `auth.StartsWith("auth0")`

source/OptimaJet.DWKit.StarterApplication/Controllers/BaseController.cs
chrisvire: I still like HttpCurrentUserContext being low-level (just above the Network (HttpContext/Auth0 Management API)).  There doesn't seem to be value in pushing it up to contain model objects.

To me the best place to FindOrCreateUser is at the BaseController.

NullVoxPopuli:
I guess merge if you are blocked. Things can always change if they need to :)

PR #48, opened 2018-07-27
Adding ProjectController

chrisvire:


source/OptimaJet.DWKit.StarterApplication/Repositories/ProjectRepository.cs
```diff
@@ -0,0 +1,84 @@
...
+            ) : base(loggerFactory, jsonApiContext, contextResolver)
+        {
+            this.OrganizationContext = organizationContext;
+            this.CurrentUserContext = currentUserContext;
+            this.UserRepository = userRepository;
+        }
+
+        public override IQueryable<Project> Get()
+        {
+            if (OrganizationContext.SpecifiedOrganizationDoesNotExist) return Enumerable.Empty<Project>().AsQueryable();
```
NullVoxPopuli: since a lot of this behavior may be duplicated between the controllers, I wonder if it'd be worth it to pull out into a BaseRepository or some other shared context that could evaluate most of this user / org logic. 

Maybe the api could be:

```cs
// always returns an array
var orgIds = GetCurrentOrganizationIds();

return base.Get()
           .Where(p => orgIds.Contains(p.OrganizationId)
```

off hand, I don't know if this would have any performance impacts, if any, but I think it could reduce the amount of code we have to maintain. :)

chrisvire: Added `User.OrganizationIds`

PR #58, opened 2018-08-02
Stub out needed things to make testing with a current user work

NullVoxPopuli:


source/OptimaJet.DWKit.StarterApplication/Services/HttpCurrentUserContext.cs
```diff
@@ -30,18 +30,24 @@ public class HttpCurrentUserContext : ICurrentUserContext
             this.TokenService = tokenService;
         }
 
-        private string AuthType {
-            get {
-                if (authType == null) {
+        private string AuthType
```
NullVoxPopuli: I found out ctrl+k, ctrl+d auto-formats in Visual Studio Community Edition
(I switched to VS Community Edition, for debugging, as I haven't figured out a good way to handle that in vim yet)

NullVoxPopuli: so, I need to figure out a good way we can get more automated styling of the C# code

NullVoxPopuli:
Next TODO will be to add backend tests to the CI runner.
right now, the ci runner does nothing for backend changes.

source/SIL.AppBuilder.Portal.Backend.Tests/Acceptance/Support/TestCUrrentUserContext.cs
```diff
@@ -0,0 +1,17 @@
+﻿using OptimaJet.DWKit.StarterApplication.Services;
```
davidmoore1: The filename is TestCUrrentUserContext according to the display here.  Minor but CU should be Cu.

NullVoxPopuli: most def. been havin Visual Studio issues with renaming today

PR #62, opened 2018-08-03
Mobile- side nav doesn't close when a link is clicked

giancorzo:
select
close

### DIFF
--- a/source/SIL.AppBuilder.Portal.Frontend/src/ui/components/sidebar/sidebar.scss
+++ b/source/SIL.AppBuilder.Portal.Frontend/src/ui/components/sidebar/sidebar.scss
@@ -24,4 +24,25 @@
   .ui.secondary.vertical.menu>.item {
     font-size: 16px;
   }
+
+   @media screen and (max-width: 480px) {
+    .item-link {
+      display: none !important;
+    }
+
+    .item-link-mobile {
+      display: block !important;
+    }
+   }
+
+   @media screen and (min-width: 480px) {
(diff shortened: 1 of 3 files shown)


source/SIL.AppBuilder.Portal.Frontend/src/ui/components/sidebar/sidebar.scss
```diff
@@ -24,4 +24,25 @@
   .ui.secondary.vertical.menu>.item {
     font-size: 16px;
   }
+
+   @media screen and (max-width: 480px) {
```
NullVoxPopuli: you should be using the built-in responsive helpers instead of these.

for example, in flex.scss, there is:

```scss
@mixin flexUtils($modifier) {
  .flex-row#{$modifier} {
    display: flex !important;
    flex-direction: row;
  }
}

// generates the above with the following modifications / breakpoints:
//  - xs
//  - sm
//  - md
//  - lg
//  - xl
@each $breakpoint in map-keys($grid-breakpoints) {
  @include media-breakpoint-up($breakpoint) {
    $modifier: '-#{$breakpoint}';

    @include flexUtils($modifier);
  }

  @include media-breakpoint-only($breakpoint) {
    @include flexUtils('-#{$breakpoint}-only');
  }
}

@include flexUtils('');
```

So, you could do the same pattern with:

````scss
.show#{$modifier} {
 // whatever in here
}
.hide#{$modifier} {

}
```

PR #64, opened 2018-08-06
Moving locale selector to header

giancorzo:
selector


NullVoxPopuli:
Is there a library out there that will provide flags for each locale?
or instead, maybe we say the full language / region? 
like, 
```
English - US
Spanish - Peru
```

PR #70, opened 2018-08-09
Update Model Attributes

chrisvire:
* Use ITrackDate to have DbContext track the DateCreated DateUpdated
  fields automatically
* Add Project.DateArchived
* Add User.ProfileVisibility (default to Public=1)
* Disable ProjectRepository.ValidateOrganization (is not working)


NullVoxPopuli:
@chrisvire looks like we have some test failures?

PR #71, opened 2018-08-09
Search + Filtering: Part 1 - "Scaffolding" and fixing the tests / making them more stable.

NullVoxPopuli:
This is part 1.
- It includes some data refactoring.
- Being more conventional in how we handle the current user data.
- Adds Search and hooks up filtering, though these requests to the backend won't yet succeed.

This PR has gotten too large, so, I'd prefer to merge this as an incremental implementation (should should be fine since we don't yet have users)

Accidentally fixes: https://developertown.visualstudio.com/SIL%20International%20Scriptoria/_workitems/edit/28387

and

https://developertown.visualstudio.com/SIL%20International%20Scriptoria/_workitems/edit/28385

Something we need to figure out is how to test that we are sending the correct headers with a particular request.

This may need to be some sort of hook into polly, or overriding `fetch` not sure yet.


source/SIL.AppBuilder.Portal.Frontend/src/data/store.ts
```diff
@@ -29,6 +29,7 @@ class CustomJSONAPISerializer extends JSONAPISerializer {
   resourceKey(type: string) { return 'remoteId'; }
 }
 
+export const serializer = new JSONAPISerializer({ schema, keyMap });
```
NullVoxPopuli: in case we ever need to normalize data outside of orbit, we can do `serializer.deserializeDocument(...)`

source/SIL.AppBuilder.Portal.Frontend/src/redux-store/provider.tsx
```diff
@@ -68,7 +68,12 @@ export default class ReduxProvider extends React.Component<IProps> {
       //       we need to manually set this specific key
       setCurrentOrganizationId(currentState.data.currentOrganizationId);
 
-      localStorage.setItem('reduxState', JSON.stringify(toPersist));
+      try {
+        localStorage.setItem('reduxState', JSON.stringify(toPersist));
+      } catch (e) {
+        console.error(toPersist, e);
```
NullVoxPopuli: if we ever try to to persist a graph cycle, we'll get an error. but we need to know what the data looked like before the stringify

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/project-search/index.tsx
```diff
@@ -0,0 +1,82 @@
...
+import { query, defaultOptions } from '@data';
+import { IFilter } from '@data/containers/with-filtering';
+import { TYPE_NAME as ORGANIZATION } from '@data/models/organization';
+import { TYPE_NAME as GROUP } from '@data/models/group';
+
+// For use after MVP and resolution of this issue:
+// https://github.com/timberio/react-query-assist/issues/25
+// import Footer from './footer';
+
+const inputProps = {
```
NullVoxPopuli: the query assist component forces us to use emotion which is css-in-js. 
https://github.com/timberio/react-query-assist/blob/master/package.json#L66

NullVoxPopuli: so, I guess if we have component-specific things that we don't want to write css for, we can use this now.

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/project-search/index.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+class Search extends React.Component<IProps> {
+  buildOptions = () => {
+    let { languages } = this.props;
+
+    languages = languages || [];
+
+    const completionData = [
+      {
+        name: 'name',
```
NullVoxPopuli: as far as I know, these are the only fields we'll be searching over.
we could add additional ones, but then UX-wise, would it even make sense to have dropdown filters?

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/project-table/data.tsx
```diff
@@ -71,6 +71,6 @@ export function withData(WrappedComponent) {
   }
 
   return compose(
-    query(mapNetworkToProps),
+    query(mapNetworkToProps, { passthroughError: true }),
```
NullVoxPopuli: later on, we may desire to make passthroughError the default, and instead add an option saying `captureError: true` for today's behavior

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/project-table/row.tsx
```diff
@@ -26,7 +26,9 @@ class Row extends React.Component<IProps & i18nProps> {
   render() {
     const { project: data, organization, t } = this.props;
     const { attributes: project } = data;
-    const { name: orgName } = organization.attributes;
+    // the organization _shouldn't_ be missing attributes
+    // but it certainly can, when fake prorject data is used elsewhere.
+    const { name: orgName } = ( organization || {} ).attributes || {};
```
giancorzo: we are no longer using attributesFor method?

source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/project/__tests__/acceptance-test.tsx
```diff
@@ -61,66 +20,95 @@ describe.only('Acceptance | Archive Project', () => {
...
-                  'date-created': "2018-08-09T20:23:54.809962",
-                  'date-updated': "2018-08-10T23:59:55.7694263Z",
-                  'date-archived': null
-                },
-                relationships: {}
-              });
-          });
+    describe('reactivating a project', () => {
+      beforeEach(function () {
+        this.mockPatch(200, 'projects/1', {
```
giancorzo: how is that the page loads if it doesn't have the other parameters?

source/SIL.AppBuilder.Portal.Frontend/src/lib/debug.tsx
```diff
@@ -0,0 +1,12 @@
+import * as React from 'react';
```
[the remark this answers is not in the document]
NullVoxPopuli: Handy for testing hoc chains in compose

source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/project/__tests__/acceptance-test.tsx
NullVoxPopuli: They end up just not being rendered. Because react doesn't render undefined values

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/project-table/row.tsx
NullVoxPopuli: forgot about it. adding.

PR #73, opened 2018-08-10
normalize env vars

chrisvire:
* Remove EmailSettings (not being used)
* Use POSTGRES_{HOST,DB,PORT,USER,PASSWORD} instead of ConnectionString:default
* API: use API_URL to get PORT
* OrgInvReq: be consistent with BaseUrl


chrisvire:
Waffling on configuration.  I am going back to API_PORT for what api uses.  It doesn't need the full url and is insignificant.  It doesn't really help being "consistent" with what nginx uses.

PR #74, opened 2018-08-13
Always allow current user to get own profile

chrisvire:
* getting users can get filtered by the current organization (which
  the user might not be in).

### DIFF
--- a/source/SIL.AppBuilder.Portal.Backend.Tests/Acceptance/APIControllers/UsersTest.cs
+++ b/source/SIL.AppBuilder.Portal.Backend.Tests/Acceptance/APIControllers/UsersTest.cs
@@ -29,10 +29,12 @@ namespace SIL.AppBuilder.Portal.Backend.Tests
         [Fact]
         public async Task Get_CurrentUser_Creates_User()
         {
+            int before = ReadTestData<AppDbContext, User>().Count;
             var user = await GetCurrentUser();
+            int after = ReadTestData<AppDbContext, User>().Count;
 
             Assert.Equal("test-auth0-id", user.ExternalId);
-            Assert.Equal(1, user.Id);
+            Assert.Equal(after, before + 1);
         }
 
         [Fact]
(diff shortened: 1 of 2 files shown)


NullVoxPopuli:
Needs a test :)

chrisvire:
@NullVoxPopuli: Actually, I had just changed the title to include "WIP" since I decided I should add a test.  Well, I should have written the failing test first... :-)

source/SIL.AppBuilder.Portal.Backend.Tests/Acceptance/APIControllers/UsersTest.cs
```diff
@@ -118,5 +120,26 @@ public async Task Patch_SomeUser_Different_Organization_NotFound()
...
+        public async Task Get_CurrentUser_Different_Organization_Found()
+        {
+            var user = await GetCurrentUser();
+
+            NeedsTestData<AppDbContext, Organization>(new List<Organization>
+            {
+                new Organization { Name = "Foo"}
+            });
+
+            var response = await Get($"/api/users/{user.Id}", "Foo");
```
NullVoxPopuli: I believe the org second arg should be an org id. Even though

chrisvire: Ah yes...

source/SIL.AppBuilder.Portal.Backend.Tests/Acceptance/APIControllers/UsersTest.cs
```diff
@@ -118,5 +120,26 @@ public async Task Patch_SomeUser_Different_Organization_NotFound()
...
+
+        /// <summary>
+        /// Getting the current user (edit the user profile) when a different
+        /// organization is the current organization (e.g. viewing users from
+        /// a different organization) should always be allowed and return the
+        /// current user.
+        /// </summary>
+        /// <returns>The current user different organization found.</returns>
+        [Fact]
+        public async Task Get_CurrentUser_Different_Organization_Found()
```
NullVoxPopuli: This test name should be more meaningful.
maybe something like `Get_User_WhenIdMatchesTheCurrentUser_OrganizationIdDoesNotMatter()` or something to that effect.

chrisvire: `Get_CurrentUser_CurrentOrganizationDoesNotMatter_Found`

NullVoxPopuli: it's not the CurrentUser endpoint though.
I guess I've (maybe) been doing

`{HTTP Verb}_{Action}_{Context?}_{Assertion}`

PR #77, opened 2018-08-14
Add OrganizationProductDefinitions

davidmoore1:


source/SIL.AppBuilder.Portal.Backend.Tests/Acceptance/APIControllers/OrganizationsTest.cs
```diff
@@ -45,5 +128,47 @@ public async Task GetOrganizations_Two()
...
+
+            Assert.Equal(HttpStatusCode.OK, response.StatusCode);
+
+            var orgs = await DeserializeList<Organization>(response);
+
+            Assert.Equal(3, orgs.Count);
+            List<Organization> objList = orgs as List<Organization>;
+            Assert.Single(objList[0].OrganizationProductDefinitions);
+            var orgProdDefinition = objList[0].OrganizationProductDefinitions[0];
+            Assert.Equal("TestProd1", orgProdDefinition.ProductDefinition.Name);
```
davidmoore1: At this point, orgProdDefinition has the ID set, but the organizationId and the productDefinitionId are both zero and the organization and product definition objects are both null

NullVoxPopuli: can you output the raw json to the log to see how things are structured in there? 

reason I ask is because the product definitions _should_ be in the `included` array at the root of the json payload, so I'm not sure that `DeserializeList` takes `included` records into account.

PR #80, opened 2018-08-15
Bugfix: Notifications

giancorzo:
Bug fix for this bug

https://developertown.visualstudio.com/SIL%20International%20Scriptoria/_workitems/edit/28328

### DIFF
--- a/source/SIL.AppBuilder.Portal.Frontend/src/ui/components/header/notifications/index.tsx
+++ b/source/SIL.AppBuilder.Portal.Frontend/src/ui/components/header/notifications/index.tsx
@@ -33,7 +33,7 @@ class Notifications extends React.Component<IProps> {
     visible: false
   };
 
-  // TODO: Remove this method we are connected to the Backend
+  // TODO: Remove this method when we are connected to the Backend
   generateRandomTaks = async () => {
     await this.props.updateStore(t => [
       t.addRecord({
@@ -64,7 +64,7 @@ class Notifications extends React.Component<IProps> {
   }
 
   componentDidMount() {
-    this.generateRandomTaks();
+    this.generateRandomTaks(); //TODO: Remove this when we get notifications from the backend
   }
 
   seenNotifications = () => {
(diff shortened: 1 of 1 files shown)


source/SIL.AppBuilder.Portal.Frontend/src/ui/components/header/notifications/index.tsx
```diff
@@ -104,11 +110,30 @@ class Notifications extends React.Component<IProps> {
...
+    if (this.state.visible) {
+      this.markNotificationsToSeen();
+    }
+
     this.setState({
       visible: !this.state.visible
     });
   }
 
+  sortNotifications = (notifications) => {
```
NullVoxPopuli: the first part of this sort doesn't do anything, and I know it was just extracted from render before, and I meant to say something earlier, but what is happening here, is that that the notifications are being sorted by time, and then resorted by isViewed, which does not consider the sort by time. 

once the second sort runs, it's possible that notifications will not be in time-order. 

ya don't need to fix it with this PR, since you just moved it from render, but just be aware of it in the future :)

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/header/notifications/index.tsx
```diff
@@ -164,10 +179,10 @@ class Notifications extends React.Component<IProps> {
                     }
 
                     return (
-                      <div
-                        className={`notification-item ${isViewed ? 'seen' : 'not-seen'}`}
-                        key={index}
-                      >
+                      <div className={`notification-item ${isViewed ? 'seen' : 'not-seen'}`} key={index} onClick={e => {
```
NullVoxPopuli: ternaries should not be used in the template portion of a component.
assigning to a variable before the template provides more semantic context. :)
Also, I've recently read that `key` shouldn't be assigned to the loop index, but something that ties the render to the item being rendered, maybe like the `id` of the notification

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/header/notifications/index.tsx
giancorzo: Will fix it later

giancorzo:
I integrated the bugfix to the refactor so I don't have to merge and update the bugfix

PR #83, opened 2018-08-16
Bugfix: Profile Timezone fix

giancorzo:
You can persist timezone now

timezon


source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/users/edit/index.tsx
```diff
@@ -24,30 +24,33 @@ export interface IOwnProps {
...
-        attributes: { ...formData, imageData }
-      }), defaultOptions());
+        attributes: { ...formData }
+      }), {
+        sources: {
+          remote: {
+            settings: {
+              headers: {
+                ...defaultHeaders(),
+                ['content-type']: 'application/vnd.api+json'
```
NullVoxPopuli: is this needed?
what happens if you replace this entire sources object with defaultOptions()?

PR #86, opened 2018-08-17
Bugfix/orgswitcher collapse after switch

NullVoxPopuli:
Fixes: https://developertown.visualstudio.com/SIL%20International%20Scriptoria/_workitems/edit/28386


source/SIL.AppBuilder.Portal.Frontend/tests/helpers/auth.ts
```diff
@@ -16,7 +16,7 @@ export function useFakeAuthentication() {
 
     server.namespace('/api', () => {
```
giancorzo: You could use mockGet to reduce code

PR #88, opened 2018-08-17
WIP Add project api for build engine

davidmoore1:
On add or update, perform validation and throw an exception
if the update is invalid

### DIFF
--- a/source/OptimaJet.DWKit.StarterApplication/Repositories/ProjectRepository.cs
+++ b/source/OptimaJet.DWKit.StarterApplication/Repositories/ProjectRepository.cs
@@ -16,6 +16,8 @@ namespace OptimaJet.DWKit.StarterApplication.Repositories
         public IOrganizationContext OrganizationContext { get; }
         public ICurrentUserContext CurrentUserContext { get; }
         public UserRepository UserRepository { get; set; }
+        public GroupRepository GroupRepository { get; set; }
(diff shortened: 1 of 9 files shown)


source/OptimaJet.DWKit.StarterApplication/Repositories/ProjectRepository.cs
```diff
@@ -48,35 +54,104 @@ public override IQueryable<Project> Get()
...
         public override async Task<Project> CreateAsync(Project entity)
         {
-            ValidateOrganization(entity);
+            ValidateOrganizationAdd(entity);
             return await base.CreateAsync(entity);
         }
 
-        private void ValidateOrganization(Project project)
+
+        private bool ValidateOranizationUpdate(Project projectUpdate, int id)
```
NullVoxPopuli: Instead of writing validation here, in the repository, I think the Form object pattern would suit this better. The form objects would then be used form the service, rather than the repository. The reason for this is that the repository should care as little about application/domain logic as possible. 

For errors, you'll also probably want to utilize jsonapi-errors rather than logging.

Some resources:

https://json-api-dotnet.github.io/#/custom-errors

I believe, according to this documentation, if you raise an exception, the jsonapi middleware will correctly capture it for you. (be sure to have tests that verify this though).
- this would clean up the code _a lot_ :)

PR #89, opened 2018-08-17
WIP: Bugfix/current user endpoint does not respect query params

NullVoxPopuli:


davidmoore1:
With the obvious comment that the test needs to pass.

NullVoxPopuli:
Yup, problem isn't solved yet :)

source/OptimaJet.DWKit.StarterApplication/Controllers/UsersController.cs
```diff
@@ -25,10 +23,10 @@ public class UsersController : BaseController<User>
         }
 
         [HttpGet("current-user")]
-        public IActionResult GetCurrentUser() {
+        public async Task<IActionResult> GetCurrentUser() {
             var currentUser = CurrentUser;
 
-            return Ok(currentUser);
+            return await base.GetAsync(currentUser.Id);
```
chrisvire: Change User.GetByAuth0Id to call DefaultGet()

```
        public async Task<User> GetByAuth0Id(string auth0Id)
            => await DefaultGet()
                .Where(e => e.ExternalId == auth0Id)
                .FirstOrDefaultAsync();
```
Then we can use currentUser here and be consistent with a User anytime it is being used (i.e. it always has the correct included members).

chrisvire: I almost made that change earlier, but now it seems the right way to solve it.

source/OptimaJet.DWKit.StarterApplication/Controllers/BaseController.cs
```diff
@@ -35,13 +36,14 @@ public User CurrentUser
         {
             get {
                 var exists = HttpContext.Items.ContainsKey(CURRENT_USER_KEY);
+                var existing = HttpContext.Items[CURRENT_USER_KEY];
 
-                if (exists) return (User)HttpContext.Items[CURRENT_USER_KEY];
+                if (exists && existing != null) return (User)existing;
```
NullVoxPopuli: even if the key exists, we want to treat a null value the same as not existing.

source/OptimaJet.DWKit.StarterApplication/Controllers/BaseController.cs
```diff
@@ -35,13 +36,14 @@ public User CurrentUser
...
 
                 // current user has not yet been found for this request.
                 // find or create because users are managed by auth0 and
                 // creation isn't proxied through the api.
-                var user = FindOrCreateCurrentUser();
+                var user = FindOrCreateCurrentUser().Result;
```
NullVoxPopuli: since async/await is more of a natural flow of async behavior, I pulled of the sync forcing up a level to here

source/OptimaJet.DWKit.StarterApplication/Controllers/UsersController.cs
NullVoxPopuli: I got rid of DefaultGet, because we want to utilize the builtin jsonapi `Get` so we can have dynamically `Included` relationships. There is a way to manually crawl through the the query params' relationships and build that query ourselves, but I didn't think it appropriate for right here.

source/OptimaJet.DWKit.StarterApplication/Repositories/UserRepository.cs
```diff
@@ -26,46 +26,47 @@ IDbContextResolver contextResolver
...
         public override IQueryable<User> Get()
         {
-            if (OrganizationContext.SpecifiedOrganizationDoesNotExist) return Enumerable.Empty<User>().AsQueryable();
+            if (OrganizationContext.SpecifiedOrganizationDoesNotExist) {
+                return Enumerable.Empty<User>().AsQueryable();
+            }
+
+            var currentUser = GetByAuth0Id(CurrentUserContext.Auth0Id).Result;
+
+            var query = base.Get();
```
NullVoxPopuli: I'm not sure if Entity Framework has an `or` query syntax, but if it did, that'd be a neat thing to add to the base query here to be shared with the two scenarios below to make sure the current-user is included in the results.

source/OptimaJet.DWKit.StarterApplication/Repositories/UserRepository.cs
```diff
@@ -26,46 +26,47 @@ IDbContextResolver contextResolver
...
-                return DefaultGet()
-                           .Where(u => u.OrganizationMemberships.Select(o => o.OrganizationId).Intersect(currentUser.OrganizationIds).Any());
-            }
-            // Get users in the current organization
-            return DefaultGet()
-                       .Where(u => u.OrganizationMemberships.Any(o => o.OrganizationId == OrganizationContext.OrganizationId));
-        }
+                // No organization specified, so include all users in the all the 
+                // organizations that the current user is a member
+                var orgIds = currentUser.OrganizationIds ?? Enumerable.Empty<int>();
```
NullVoxPopuli: in the event a user has no memberships, we previously would get a 500 error.

source/OptimaJet.DWKit.StarterApplication/Repositories/UserRepository.cs
```diff
@@ -26,46 +26,47 @@ IDbContextResolver contextResolver
...
 
-        public override async Task<User> GetAsync(int id)
-        {
-            // The default implementation filters by selected organization.  The
-            // current user might not be in that organization.
-            // Always allow getting the current user.
-            var currentUser = await GetByAuth0Id(CurrentUserContext.Auth0Id);
-            if (currentUser != null && currentUser.Id == id) {
-                return await DefaultGet().Where(e => e.Id == id).FirstAsync();
+                return query
```
NullVoxPopuli: some formatting tweaks, as I don't like horizontal scrolling :)

source/OptimaJet.DWKit.StarterApplication/Repositories/UserRepository.cs
```diff
@@ -26,46 +26,47 @@ IDbContextResolver contextResolver
...
+                        )
+                ));
         }
 
         public async Task<User> GetByAuth0Id(string auth0Id)
             => await base.Get()
                 .Where(e => e.ExternalId == auth0Id)
-                .Include(u => u.OrganizationMemberships)
-                    .ThenInclude(om => om.Organization)
+                .Include(user => user.OrganizationMemberships)
```
NullVoxPopuli: only the OrganizationMembership is included here, because it is used above in the `Intersect`. It's also how I learned about the possible 500 exception when there are no OrganizationMemberships (which was the behavior when this Include was omitted).

The Organization is not needed anywhere as far as I could tell. (at least no tests create a situation where it needs to be loaded)

source/SIL.AppBuilder.Portal.Backend.Tests/Acceptance/APIControllers/UsersTest.cs
```diff
@@ -26,6 +26,24 @@ private async Task<User> GetCurrentUser()
             return user;
         }
 
+        [Fact]
+        public async Task Get_CurrentUser_Includes_IncludedResources()
+        {
+            NeedsConfiguredCurrentUser();
+
+            var response = await Get("/api/users/current-user?include=organization-memberships");
```
NullVoxPopuli: the include parameter on the query-side can now be dynamic, and we don't need to manually include the world on the backend.

source/OptimaJet.DWKit.StarterApplication/Repositories/UserRepository.cs
```diff
@@ -26,46 +26,47 @@ IDbContextResolver contextResolver
...
-            }
-            // Get users in the current organization
-            return DefaultGet()
-                       .Where(u => u.OrganizationMemberships.Any(o => o.OrganizationId == OrganizationContext.OrganizationId));
-        }
+                // No organization specified, so include all users in the all the 
+                // organizations that the current user is a member
+                var orgIds = currentUser.OrganizationIds ?? Enumerable.Empty<int>();
 
-        public override async Task<User> GetAsync(int id)
```
NullVoxPopuli: I got rid of GetAsync(id), because Its implementation relies on the Get(), so I moved the inclusion of the current-user into there.

source/OptimaJet.DWKit.StarterApplication/Repositories/UserRepository.cs
```diff
@@ -26,46 +27,69 @@ IDbContextResolver contextResolver
             this.CurrentUserContext = currentUserContext;
         }
 
-        protected IQueryable<User> DefaultGet()
+        public override IQueryable<User> Filter(IQueryable<User> query, FilterQuery filterQuery)
```
chrisvire: Please add an explanation of why `Get()` always return the `CurrentUser` if if the current user isn't in the specified organization (using the `Organization: orgId` header) and that the user of the api will still need to add a `filter[organization-id]` to keep from getting the CurrentUser in the results.

PR #90, opened 2018-08-20
Change disable user to positive active user

giancorzo:


source/SIL.AppBuilder.Portal.Frontend/src/ui/components/user-table/data.tsx
```diff
@@ -20,18 +22,12 @@ function mapNetworkToProps(passedProps) {
     //       is resolved
     users: [
       q => q.findRecords(USER),
-      {
-        sources: {
-          remote: {
-            settings: {
-              ...defaultSourceOptions
-            },
-            include: [MEMBERSHIPS]
```
NullVoxPopuli: are the memberships not needed anymore?

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/user-table/data.tsx
```diff
@@ -70,11 +67,27 @@ export function withData(WrappedComponent) {
...
     toggleLock = async (user) => {
 
-      const { updateStore } = this.props;
+      const { updateStore, t } = this.props;
+
+      const successMessage = !user.attributes.isLocked ?
+        t('users.operations.lock.success') :
+        t('users.operations.unlock.success');
+
+      const errorMessage = !user.attributes.isLocked ?
```
NullVoxPopuli: the logic is weird (generally, ternaries shouldn't be negative (positive thoughts!))

The way I read this:

 - If the user is _not_ locked,
   - the error message is _lock_.error
 - otherwise, if the user is _locked_,
   - the error message is the _unlocked_ error

just seems like this is semantically confusing / not matching?

source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/users/list/__tests__/acceptance-test.tsx
```diff
@@ -0,0 +1,81 @@
...
+    });
+
+    it('is in users page', () => {
+      expect(location().pathname).to.equal('/users');
+    });
+
+    describe('an active users exists', () => {
+
+      it('active user', () => {
+        console.log(page.unlockUser);
```
NullVoxPopuli: stray console

source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/users/list/__tests__/acceptance-test.tsx
```diff
@@ -0,0 +1,81 @@
...
+          });
+        });
+
+        describe('toggle is clicked', () => {
+          beforeEach(async () => {
+            await page.clickLockUser();
+          });
+
+          it("user locked", () => {
+            console.log(page.unlockUser);
```
NullVoxPopuli: stray console

source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/users/list/__tests__/page.ts
```diff
@@ -0,0 +1,16 @@
...
+  interactor, hasClass,
+  clickable
+} from '@bigtest/interactor';
+
+@interactor
+export class UserInteractor {
+  constructor(selector?: string) { }
+
+  clickLockUser = clickable('[data-test-toggle-lock]:first-child');
+  unlockUser = hasClass('[data-test-toggle-lock]:first-child','checked');
```
NullVoxPopuli: this property name is a command, but the value represents a status. So, the property name should be updated to represent the value :)

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/user-table/data.tsx
giancorzo: I don't see that we are using them but I guess it would be useful later when filtering

NullVoxPopuli: they'll be needed if the users table is going to have any information about what orgs a user is a member of

NullVoxPopuli:
@giancorzo there is a conflict -- sorry. had to resolve something for Liz.

PR #91, opened 2018-08-21
Execute migrations in Production Image

chrisvire:


source/OptimaJet.DWKit.StarterApplication/Startup.cs
```diff
@@ -103,11 +103,11 @@ public virtual void ConfigureDatabase(IServiceCollection services, string name =
...
         {
-            var dbHost = GetVarOrThrow("POSTGRES_HOST");
+            var dbHost = GetVarOrDefault("POSTGRES_HOST", "localhost");
             var dbPort = GetVarOrDefault("POSTGRES_PORT", "5432");
-            var dbDatabase = GetVarOrThrow("POSTGRES_DB");
-            var dbUser = GetVarOrThrow("POSTGRES_USER");
-            var dbPassword = GetVarOrThrow("POSTGRES_PASSWORD");
+            var dbDatabase = GetVarOrDefault("POSTGRES_DB", "development");
+            var dbUser = GetVarOrDefault("POSTGRES_USER", "db-user");
+            var dbPassword = GetVarOrDefault("POSTGRES_PASSWORD", "1234");
```
chrisvire: These are needed so that the `dotnet ef migrations script --idempotent` command can run during build-time.

NullVoxPopuli: I mean, the .env file is typically source'd into the current shell environment before running. :-\

chrisvire: During building of the Docker image, there is no .env file, right?

NullVoxPopuli: you can load .env with bash. 

```
set -a
source .env
dotnet ...
```

README.md
```diff
@@ -24,6 +24,27 @@ docker run -p 8080:80 nginx-$CURRENT_VERSION
 docker run -p 3000:7081 api-$CURRENT_VERSION
 ```
 
+## Production
```
NullVoxPopuli: :clap: :clap: :clap:

chrisvire:
@NullVoxPopuli OK, I think I have taken care of your concerns.

source/Dockerfile.backend
```diff
@@ -26,8 +26,20 @@ COPY ./ /app/
 COPY --from=dependencies /app /app
 WORKDIR /app/OptimaJet.DWKit.StarterApplication
 RUN dotnet publish -c Release -o out
+# Need dummy values for POSTGRES_* variables for migrations to run
+RUN POSTGRES_HOST=build \
```
NullVoxPopuli: I don't understand this comment.
what do you mean?

I thought the variables were checked at runtime? in which case you'd want to set them in the deployment environment:
like, 
image

in kubernetes.

or like, 
```
docker run --env VAR1=value1 --env VAR2=value2 ubuntu 
```

using vanilla docker

chrisvire: `dotnet ef migrations` loads the assembly (using its own main) and dynamically calls `public static IWebHost BuildWebHost(string[] args)` to acquire the DbContext to run the migrations.  This all happens at build-time.

Our code expects to be run only during run-time and requires (via `GetVarOrThrow`) values for POSTGRES_*.

A database connection is not required for generation of the migrations at build-time, so this change provides some dummy values to satisfy our startup code in `Startup.ConfigureServices`.

PR #92, opened 2018-08-21
Creating a Project

NullVoxPopuli:
https://developertown.visualstudio.com/SIL%20International%20Scriptoria/_workitems/edit/26309

- [x] **A project cannot be created if no group assigned**
  - [x] Frontend
  - [ ] Backend
- [x] **A user can create a project for their group and cannot create a project for a group in which they don't belong**
  - [x] Frontend - show only groups the user is a member of
    - [x] If the user has no groups, don't allow them to even see the create project screen.
  - [ ] Backend
- [x] **Project title, type (SAB, RAB), language code fields are required**
  - [x] Frontend
  - [ ] Backend
- [x] **Public setting toggle should be visible, defaulted on in positive language**
  - [ ] Backend needs a visibility field to persist the visibility
- [x] **Request approval button should be changed to say "save" and data should save (for now since workflow isn't complete-- "request approval" is a separate story (28402)**
- [x] Frontend Acceptance Tests


source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/projects/new/with-access-restriction.tsx
```diff
@@ -0,0 +1,21 @@
+import * as React from 'react';
+import { Redirect } from 'react-router-dom';
+
+import * as toast from '@lib/toast';
+import { hasRelationship } from '@data';
+
+export function withAccessRestriction(WrappedComponent) {
```
giancorzo: name of the HOC doesn't reflect what it does

PR #94, opened 2018-08-22
Upgrade to DWKit 2.2

chrisvire:


chrisvire:
I will make a separate PR to fix all the Optimajet namespace stuff in Model classes.

PR #97, opened 2018-08-23
Bugfix: Profile wasn't saving

giancorzo:
I changed to defaultOptions and now it's saving correctly except for email notification that we don't have it yet in the backend


NullVoxPopuli:
was the issue that the content type was specified twice in the HTTP header?

giancorzo:
Yeah we where setting the content type twice, that was the error

PR #99, opened 2018-08-23
Handling 403 forbidden (bug 28489)

giancorzo:
I added a case to handle forbidden users

### DIFF
--- a/source/SIL.AppBuilder.Portal.Frontend/src/data/containers/with-current-user.tsx
+++ b/source/SIL.AppBuilder.Portal.Frontend/src/data/containers/with-current-user.tsx
@@ -2,11 +2,8 @@ import * as React from 'react';
 import { Redirect } from 'react-router-dom';
 import { withData, WithDataProps } from 'react-orbitjs';
 import { compose } from 'recompose';
-import Orbit from '@orbit/data';
 
 import { defaultSourceOptions, pushPayload } from '@data';
-import  { keyMap } from '@data/schema';
-import { serializer } from '@data/store';
 import { UserAttributes, TYPE_NAME } from '@data/models/user';
 
 import { getAuth0Id } from '@lib/auth0';
@@ -16,6 +13,8 @@ import { hasRelationship } from '@data';
 import PageLoader from '@ui/components/loaders/page';
 import OrgMembershipRequired from '@ui/components/errors/org-membership-required';
 import PageError from '@ui/components/errors/page';
+
+import { deleteToken } from '@lib/auth0';
 
 type UserPayload = JSONAPIDocument<UserAttributes>;
 
@@ -79,10 +78,14 @@ export function withCurrentUser() {
         this.fetchCurrentUser();
       }
 
+      isUserLocked = (response) => {
+        return response.status === 403;
+      }
+
       fetchCurrentUser = async () => {
         const { networkFetchComplete, currentUser } = this.state;
 
-        const { updateStore, queryStore, usersMatchingLoggedInUser: fromCache } = this.props;
+        const { updateStore, usersMatchingLoggedInUser: fromCache } = this.props;
         const auth0IdFromJWT = getAuth0Id();
 
         // if we have a currentUser from cache, and the currentUser
@@ -107,6 +110,12 @@ export function withCurrentUser() {
           // https://github.com/json-api-dotnet/JsonApiDotNetCore/issues/39
           // is resolved
           const response = await authenticatedGet('/api/users/current-user?include=organization-memberships,group-memberships');
+
+          if (this.isUserLocked(response)) {
+            console.debug('Current user is Locked');
+            throw new Error('Your user has been locked, contact you organization administrator');
+          }
+
           const json = await tryParseJson(response);
 
           await pushPayload(updateStore, json);
@@ -128,12 +137,16 @@ export function withCurrentUser() {
         const { error, currentUser, isLoading } = this.state;
 
         if (error) {
+          debugger;
           return <PageError error={error} />;
         }
 
         if (isLoading) {
+
           return <PageLoader />;
+
         } else if (currentUser) {
+
           const hasMembership = hasRelationship(currentUser, 'organizationMemberships');
 
           if (hasMembership) {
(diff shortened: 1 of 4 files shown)


source/SIL.AppBuilder.Portal.Frontend/src/data/containers/with-current-user.tsx
```diff
@@ -79,10 +78,14 @@ export function withCurrentUser() {
         this.fetchCurrentUser();
       }
 
+      isUserLocked = (response) => {
```
NullVoxPopuli: 403 won't always mean the user is locked. so I don't think it makes sense to always extract this to a method

source/SIL.AppBuilder.Portal.Frontend/src/data/containers/with-current-user.tsx
```diff
@@ -107,6 +110,12 @@ export function withCurrentUser() {
           // https://github.com/json-api-dotnet/JsonApiDotNetCore/issues/39
           // is resolved
           const response = await authenticatedGet('/api/users/current-user?include=organization-memberships,group-memberships');
+
+          if (this.isUserLocked(response)) {
+            console.debug('Current user is Locked');
+            throw new Error('Your user has been locked, contact you organization administrator');
```
NullVoxPopuli: why are you throwing an error?

source/SIL.AppBuilder.Portal.Frontend/src/data/containers/with-current-user.tsx
```diff
@@ -128,12 +137,16 @@ export function withCurrentUser() {
         const { error, currentUser, isLoading } = this.state;
 
         if (error) {
+          debugger;
           return <PageError error={error} />;
         }
 
         if (isLoading) {
+
```
NullVoxPopuli: why are you adding so many blank lines everywhere?

source/SIL.AppBuilder.Portal.Frontend/src/lib/fetch.ts
```diff
@@ -2,6 +2,7 @@ import { getToken } from './auth0';
 import { getCurrentOrganizationId } from './current-organization';
 
 export const tryParseJson = async (response) => {
+
```
NullVoxPopuli: ??? also unneeded change

source/SIL.AppBuilder.Portal.Frontend/src/data/containers/with-current-user.tsx
giancorzo: So if 403 is not the correct way to check if a user is Blocked should I try to read the error message and see if has "some keyword about blocked" ?

source/SIL.AppBuilder.Portal.Frontend/src/data/containers/with-current-user.tsx
giancorzo: I'm throwing an error so I can reuse the error page

        if (error) {
          return <PageError error={error} />;
        }

source/SIL.AppBuilder.Portal.Frontend/src/lib/fetch.ts
giancorzo: It's easier to read if we always add 1 line at the beginning of the function

NullVoxPopuli: I disagree, if there were a rule for this, I'd add it. :-p
But in general, try not to make changes in your PRs that don't directly affect your PRs.

source/SIL.AppBuilder.Portal.Frontend/src/data/containers/with-current-user.tsx
NullVoxPopuli: checking for 403 is correct, but 403 will not always mean the user is locked

403 means forbidden

source/SIL.AppBuilder.Portal.Frontend/src/data/containers/with-current-user.tsx
NullVoxPopuli: :+1:

source/SIL.AppBuilder.Portal.Frontend/src/lib/fetch.ts
giancorzo: ok will change it, I used to follow some rules from HAPI community

source/SIL.AppBuilder.Portal.Frontend/src/lib/fetch.ts
```diff
@@ -2,8 +2,8 @@ import { getToken } from './auth0';
 import { getCurrentOrganizationId } from './current-organization';
 
 export const tryParseJson = async (response) => {
-  const text = await response.text();
 
+  const text = await response.text();
```
NullVoxPopuli: what's with this change?

NullVoxPopuli: I can't tell what's different? lol

NullVoxPopuli: oh, it changed lines? weird

giancorzo: I didn't change this one... just added some debugger and space

source/SIL.AppBuilder.Portal.Frontend/src/data/containers/with-current-user.tsx
```diff
@@ -107,6 +106,12 @@ export function withCurrentUser() {
           // https://github.com/json-api-dotnet/JsonApiDotNetCore/issues/39
           // is resolved
           const response = await authenticatedGet('/api/users/current-user?include=organization-memberships,group-memberships');
+
+          if (response.status === 403) {
+            console.debug('Current user is Forbidden');
+            throw new Error('An error occurred: Forbidden. Please contact your organization administrator.');
```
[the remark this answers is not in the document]
giancorzo: Translation added

PR #96, opened 2018-08-23
Feature/assign a group to aproject

NullVoxPopuli:
Note: this depends on the work done in https://github.com/sillsdev/appbuilder-portal/pull/92 

 - [x] **Available group options for that organization are displayed**
     - Note: this is a subset of groups available to the user
     - Note: groups shown should be (groups the user is in for the project organization) + (the current group the project is in)
 - [x] **If user is only in one group cannot select a different group (but still display what the project's group name)**
 - [x] **A project cannot be in multiple groups**
   - This is a consequence of our data model / design
 - [x] **Project to group assignment by default will be based on the group the app builder belongs to (if multiple, pick the first one)**
   - Note, this is implemented in PR #92
 - [x] Acceptance Tests

I updated some of the test stuff: 
image

image

Blockers / Questions:
 - For "Available group options for organization are displayed?"
   - is this all groups in the org?
   - would someone who isn't in the group the project is assigned to be able to see this list?
   - Note: anyone from a different organization gets redirected, so non issue there

PR #100, opened 2018-08-24
Feature/assign a user to a project

NullVoxPopuli:
Depends on https://github.com/sillsdev/appbuilder-portal/pull/96

https://developertown.visualstudio.com/SIL%20International%20Scriptoria/_workitems/edit/25967

 - [x] Project cannot be reassigned to a user that is not in the app builders group
 - [x] Visual indication - project now appears in app builders' "my projects" view

### DIFF
--- a/source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/project/__tests__/archive-acceptance-test.tsx
+++ b/source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/project/__tests__/archive-acceptance-test.tsx
@@ -7,7 +7,7 @@ import { setupApplicationTest, setupRequestInterceptor, useFakeAuthentication } 
 
 import page from './page';
 
-describe('Acceptance | Archive Project', () => {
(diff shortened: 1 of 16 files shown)


source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/project/__tests__/archive-acceptance-test.tsx
```diff
@@ -24,6 +24,7 @@ describe('Acceptance | Archive Project', () => {
 
   describe('an active project exists', () => {
     beforeEach(function() {
+      this.mockGet(200, 'users', { data: [] });
```
giancorzo: should not start with /users?

NullVoxPopuli: turns out it's not required, because mockGet defines a namespace of `api` rather than doing string interpolation

PR #101, opened 2018-08-24
Move Organization Header check to service for group user and project 

davidmoore1:
Moved organization check to service for group and User
Completed implementation for group


source/OptimaJet.DWKit.StarterApplication/Repositories/GroupRepository.cs
```diff
@@ -9,24 +11,57 @@ namespace OptimaJet.DWKit.StarterApplication.Repositories
...
+            ICurrentUserContext currentUserContext,
             IDbContextResolver contextResolver
             ) : base(loggerFactory, jsonApiContext, contextResolver)
         {
-            this.OrganizationContext = organizationContext;
+            this.OrganizationContext = organizationContext;
+            this.UserRepository = userRepository;
+            this.CurrentUserContext = currentUserContext;
+        }
+        public override IQueryable<Group> Filter(IQueryable<Group> query, FilterQuery filterQuery)
```
NullVoxPopuli: it may make sense to pull these two methods into to the base class

NullVoxPopuli: though, not if it moves into the service -- I should keep looking over things

davidmoore1:
Organization validation checks for User, Project and Group moved to Service with filters in repositories.
ValidateProject in ProjectRepository moved to Forms/Projects/CreateForm and UpdateForm
Some validation that was needed for Group Update was also added to a new UpdateForm
Tests for all new features were added.  A previous Project test was merged into the new test files.

source/OptimaJet.DWKit.StarterApplication/Controllers/BaseController.cs
```diff
@@ -67,7 +67,7 @@ private async Task<User> FindOrCreateCurrentUser()
                 FamilyName = currentUserContext.FamilyName
             };
 
-            var newEntity = await userService.CreateAsync(newUser);
+            var newEntity = await userService.EntityRepository.CreateAsync(newUser);
```
NullVoxPopuli: in general, controllers should not touch repositories

source/OptimaJet.DWKit.StarterApplication/Forms/Groups/UpdateForm.cs
```diff
@@ -0,0 +1,49 @@
...
+            GroupRepository groupRepository,
+            ICurrentUserContext currentUserContext)
+        {
+            UserRepository = userRepository;
+            GroupRepository = groupRepository;
+            CurrentUserContext = currentUserContext;
+         }
+        public bool IsValid(int id, Group group)
+        {
+            var retVal = true;
```
NullVoxPopuli: this is kind of unnecessary abbreviation, yeah? could just be `result`

source/OptimaJet.DWKit.StarterApplication/Forms/Groups/UpdateForm.cs
```diff
@@ -0,0 +1,49 @@
...
+            //If changing organization, validate the change
+            if (group.OwnerId > 0)
```
NullVoxPopuli: the comment and the code don't say the same thing?

this is just checking that the group _has_ an owner? maybe? I guess in what situation would the ownerId `== 0`?

source/OptimaJet.DWKit.StarterApplication/Forms/Groups/UpdateForm.cs
```diff
@@ -0,0 +1,49 @@
...
+            {
+                var currentUser = UserRepository.GetByAuth0Id(CurrentUserContext.Auth0Id).Result;
```
NullVoxPopuli: the current user should already exist in the CurrentUserContext, no need to trigger another query

source/OptimaJet.DWKit.StarterApplication/Forms/Groups/UpdateForm.cs
```diff
@@ -0,0 +1,49 @@
...
+                var orgIds = currentUser.OrganizationIds ?? Enumerable.Empty<int>();
+                if (!orgIds.Contains(group.OwnerId))
+                {
+                    retVal = false;
+                    var original = GroupRepository.Get()
+                                                  .Where(g => g.Id == id)
+                                                  .Include(g => g.Owner)
+                                                  .FirstOrDefaultAsync().Result;
+                    var message = $"Project '{original.Name}': Current user not a member of project organization '{original.Owner.Name}'";
+                    var error = new Error(422, "Project Validation Failure", message);
```
NullVoxPopuli: The error says "Project Validation failure", but this is a "Group" Update form.

source/OptimaJet.DWKit.StarterApplication/Forms/Groups/UpdateForm.cs
```diff
@@ -0,0 +1,49 @@
...
+                var currentUser = UserRepository.GetByAuth0Id(CurrentUserContext.Auth0Id).Result;
+                var orgIds = currentUser.OrganizationIds ?? Enumerable.Empty<int>();
+                if (!orgIds.Contains(group.OwnerId))
+                {
+                    retVal = false;
+                    var original = GroupRepository.Get()
+                                                  .Where(g => g.Id == id)
+                                                  .Include(g => g.Owner)
+                                                  .FirstOrDefaultAsync().Result;
+                    var message = $"Project '{original.Name}': Current user not a member of project organization '{original.Owner.Name}'";
```
NullVoxPopuli: copy-paste error?

source/OptimaJet.DWKit.StarterApplication/Forms/Groups/UpdateForm.cs
```diff
@@ -0,0 +1,49 @@
...
+                    var error = new Error(422, "Project Validation Failure", message);
+                    Errors.Add(error);
+                }
+            }
+
+            return retVal;
```
NullVoxPopuli: would it make more sense to just `return Errors.Count() == 0;`?

source/OptimaJet.DWKit.StarterApplication/Forms/Projects/CreateForm.cs
```diff
@@ -0,0 +1,94 @@
...
+                    // Allowing it to return in these cases should cause the base to hit normal
+                    // foreign key failures
+                    var message = $"Project '{project.Name}': Organization Group or Owner not present or invalid";
+                    Log.Error(message);
+                    return true;
+                }
+                if (organization != group.Owner)
+                {
+                    retVal = false;
+                    var message = $"Project '{project.Name}': Group '{group.Name}' not owned by project organization '{organization.Name}'";
```
NullVoxPopuli: all this error construction logic should be extracted to the BaseForm for simpler usage.

source/OptimaJet.DWKit.StarterApplication/Forms/Projects/CreateForm.cs
```diff
@@ -0,0 +1,94 @@
...
+                    Log.Error(message);
```
NullVoxPopuli: remove the logs. We have error objects :)

source/OptimaJet.DWKit.StarterApplication/Forms/Projects/CreateForm.cs
```diff
@@ -0,0 +1,94 @@
...
+                if (!orgIds.Contains(organization.Id))
+                {
+                    retVal = false;
+                    var message = $"Project '{project.Name}': Current user not a member of project organization '{organization.Name}'";
+                    var error = new Error(422, "Project Validation Failure", message);
+                    Errors.Add(error);
+                }
+ 
+            }
+            return retVal;
```
NullVoxPopuli: same not here about checking the errors object.
maybe having the superclass do the comparison, so at the end of each of these, you can just do `base.IsValid()`

source/OptimaJet.DWKit.StarterApplication/Forms/Projects/CreateForm.cs
```diff
@@ -0,0 +1,94 @@
...
+                }
+                if ((owner.OrganizationIds == null) || (!owner.OrganizationIds.Contains(organization.Id)))
+                {
+                    retVal = false;
+                    var message = $"Project '{project.Name}': Owner '{owner.Name}' not a member of project organization '{organization.Name}'";
+                    var error = new Error(422, "Project Validation Failure", message);
+                    Errors.Add(error);
+                }
+                // The current user should be a member of the organization
+                var currentUser = UserRepository.GetByAuth0Id(CurrentUserContext.Auth0Id).Result;
```
NullVoxPopuli: same note here about the current user

source/OptimaJet.DWKit.StarterApplication/Forms/Projects/UpdateForm.cs
```diff
@@ -0,0 +1,127 @@
...
+                || (project.GroupId != 0))
+            {
+                var original = ProjectRepository.Get()
+                    .Where(p => p.Id == id)
+                    .Include(p => p.Organization)
+                    .Include(p => p.Owner)
+                    .Include(p => p.Group)
+                       .ThenInclude(g => g.Owner)
+                    .FirstOrDefaultAsync().Result;
+                var owner = UserRepository.Get()
```
NullVoxPopuli: you should already have access to the owner from `original.Owner`? why make this extra query?

source/OptimaJet.DWKit.StarterApplication/Controllers/BaseController.cs
davidmoore1: Chris had wanted to make CreateAsync at the service level return not implemented, which is why the change was made.  Changed this to call a new DirectCreateAsync at the service level which simply calls the service base.GetAsync(id)

source/OptimaJet.DWKit.StarterApplication/Forms/Groups/UpdateForm.cs
davidmoore1: refactored to result

source/OptimaJet.DWKit.StarterApplication/Forms/Groups/UpdateForm.cs
davidmoore1: Changed all of the zeros everywhere to VALUE_NOT_SET (which is a static set to 0).  Changed this if to != the same as the others

source/OptimaJet.DWKit.StarterApplication/Forms/Projects/CreateForm.cs
davidmoore1: I have three different forms with three different setups.  I'm not exactly sure what one would extract.  I could create routines to retrieve an organization or a group or a project and call them as needed but I don't think that really helps significantly.

source/OptimaJet.DWKit.StarterApplication/Forms/Projects/CreateForm.cs
davidmoore1: Removed logs

source/OptimaJet.DWKit.StarterApplication/Forms/Groups/UpdateForm.cs
davidmoore1: Current user is not accessible in the current user context.  And while CurrentUser is available at the Controller level, it is not easily accessible at the service level where the forms are called.

source/OptimaJet.DWKit.StarterApplication/Forms/Groups/UpdateForm.cs
davidmoore1: A couple of problems within the message and error referring to project when this is group.  These have been updated.

source/OptimaJet.DWKit.StarterApplication/Forms/Projects/CreateForm.cs
davidmoore1: BaseForm isn't really the correct place because the checks are only common for the project forms.  Created a new BaseProjectForm and changed both the project forms to derive from that

source/OptimaJet.DWKit.StarterApplication/Forms/Projects/CreateForm.cs
davidmoore1: Fixed as above

source/OptimaJet.DWKit.StarterApplication/Forms/Groups/UpdateForm.cs
NullVoxPopuli: You could pass it down to the service? maybe? idk

source/OptimaJet.DWKit.StarterApplication/Forms/Projects/CreateForm.cs
NullVoxPopuli: I just meant that maybe it would make sense to have something like:

```csharp
AddError(message)
```

where `AddError` is defined on the base class as:

```
void AddError(string message) {
  var error = new Error(message);

  Errors.add(error);
}
```

or something like that -- just to remove some of the boilerplate

source/OptimaJet.DWKit.StarterApplication/Forms/Projects/UpdateForm.cs
davidmoore1: Rewrote the query so that it was possible to get all of the group membership and organization membership includes in one query

source/OptimaJet.DWKit.StarterApplication/Controllers/BaseController.cs
NullVoxPopuli: it should be at the controller level for the user. :-\

NullVoxPopuli: semantically, it should be a 405 status code, I think

source/OptimaJet.DWKit.StarterApplication/Forms/Groups/UpdateForm.cs
NullVoxPopuli: has it been pushed?

source/OptimaJet.DWKit.StarterApplication/Forms/Projects/CreateForm.cs
davidmoore1: Added that as well

source/OptimaJet.DWKit.StarterApplication/Controllers/BaseController.cs
davidmoore1: Moved change to controller and return a 405 on post to user.  Reversed previous changes so that the base controller calls user service createasync as it did originally.  Added test that verifies failure 405 adding new user.

source/OptimaJet.DWKit.StarterApplication/Services/GroupService.cs
```diff
@@ -0,0 +1,78 @@
...
+                query.Filters.Add(new JsonApiDotNetCore.Internal.Query.FilterQuery("organization-header", value, "="));
+                return await base.GetAsync();
+            }
+        }
+        public override async Task<Group> GetAsync(int id)
+        {
+            var groups = await GetAsync();
+            return groups.SingleOrDefault(g => g.Id == id);
+        }
+        public override async Task<Group> UpdateAsync(int id, Group resource)
```
NullVoxPopuli: form for update should have a new instance created in here.

source/OptimaJet.DWKit.StarterApplication/Services/ProjectService.cs
```diff
@@ -0,0 +1,92 @@
...
+            ILoggerFactory loggerFactory) : base(jsonApiContext, projectRepository, loggerFactory)
+        {
+            OrganizationContext = organizationContext;
+            JsonApiContext = jsonApiContext;
+            CurrentUserContext = currentUserContext;
+            UserRepository = userRepository;
+            CreateForm = createForm;
+            UpdateForm = updateForm;
+        }
+        public override async Task<IEnumerable<Project>> GetAsync()
```
NullVoxPopuli: there seems to be quite a bit of duplication here with the groups service.

This could probably be abstracted somewhere :)

source/OptimaJet.DWKit.StarterApplication/Services/UserService.cs
```diff
@@ -11,14 +14,60 @@ namespace OptimaJet.DWKit.StarterApplication.Services
...
+        {
+            var currentUser = EntityRepository.GetByAuth0Id(CurrentUserContext.Auth0Id).Result;
+            if (currentUser.Id == id)
+            {
+                return await base.GetAsync(id);
+            }
+            var users = await GetAsync();
+            return users.SingleOrDefault(u => u.Id == id);
+        }
+        public override async Task<User> UpdateAsync(int id, User resource)
```
NullVoxPopuli: this looks like it should almost by default behavior, yeah?

source/OptimaJet.DWKit.StarterApplication/Services/UserService.cs
```diff
@@ -11,14 +14,60 @@ namespace OptimaJet.DWKit.StarterApplication.Services
...
+        {
+            var user = await GetAsync(id);
+            if (user == null)
```
NullVoxPopuli: would we want to return null when a user isn't found? or would we want to tell the current user that the user wasn't found? (404)

source/SIL.AppBuilder.Portal.Backend.Tests/Acceptance/APIControllers/Groups/GetGroupTest.cs
```diff
@@ -0,0 +1,152 @@
...
+            });
+        }
+
+        [Fact]
+        public async Task Get_Group_With_An_OrganizationHeader()
+        {
+            BuildTestData();
+
+            var url = "/api/groups/" + group1.Id.ToString();
+            var response = await Get(url, org1.Id.ToString());
```
NullVoxPopuli: I wonder if `Get` should be able to take an Integer and automatically call `.ToString()` for you

source/SIL.AppBuilder.Portal.Backend.Tests/Acceptance/APIControllers/Groups/GetGroupTest.cs
```diff
@@ -0,0 +1,152 @@
...
+
+            Assert.Equal(HttpStatusCode.OK, response.StatusCode);
+
+            var group = await Deserialize<Group>(response);
+
+            Assert.Equal(group.Id, group1.Id);
+
+        }
+        [Fact]
+        public async Task Get_Invalid_Group()
```
NullVoxPopuli: why is the group invalid?

source/SIL.AppBuilder.Portal.Backend.Tests/Acceptance/APIControllers/Groups/GetGroupTest.cs
```diff
@@ -0,0 +1,152 @@
...
+            var response = await Get(url, org1.Id.ToString());
+
+            Assert.Equal(HttpStatusCode.OK, response.StatusCode);
+
+            var group = await Deserialize<Group>(response);
+
+            Assert.Equal(group.Id, group1.Id);
+        }
+        [Fact]
+        public async Task Get_Group_By_Id_Empty()
```
NullVoxPopuli: what's empty?

source/SIL.AppBuilder.Portal.Backend.Tests/Acceptance/APIControllers/Groups/GetGroupsTest.cs
```diff
@@ -0,0 +1,159 @@
...
+            var groups = await DeserializeList<Group>(response);
+
+            Assert.Equal(3, groups.Count);
+
+            var ids = groups.Select(g => g.Id);
+
+            Assert.Contains(group1.Id, ids);
+            Assert.Contains(group2.Id, ids);
+            Assert.Contains(group3.Id, ids);
+            Assert.DoesNotContain(group4.Id, ids);
```
NullVoxPopuli: Why is this one not included?

source/SIL.AppBuilder.Portal.Backend.Tests/Acceptance/APIControllers/Groups/GetGroupsTest.cs
```diff
@@ -0,0 +1,159 @@
...
+            Assert.Contains(group1.Id, ids);
+            Assert.Contains(group2.Id, ids);
+            Assert.DoesNotContain(group3.Id, ids);
+            Assert.DoesNotContain(group4.Id, ids);
+        }
+        [Fact] async Task Get_With_No_Organization()
+        {
+            BuildTestData();
+            var url = "/api/groups";
+            var response = await Get(url); // Empty string for org header
```
NullVoxPopuli: empty string? I see no empty string

source/SIL.AppBuilder.Portal.Backend.Tests/Acceptance/APIControllers/Groups/GetGroupsTest.cs
```diff
@@ -0,0 +1,159 @@
...
+            var ids = groups.Select(g => g.Id);
+
+            Assert.Contains(group1.Id, ids);
+            Assert.Contains(group2.Id, ids);
+            Assert.Contains(group3.Id, ids);
+            Assert.DoesNotContain(group4.Id, ids);
+
+        }
+        [Fact]
+        async Task Get_With_Invalid_Organization()
```
NullVoxPopuli: why is the organization invalid?

source/OptimaJet.DWKit.StarterApplication/Forms/Groups/UpdateForm.cs
```diff
@@ -0,0 +1,48 @@
...
+        {
+            UserRepository = userRepository;
+            GroupRepository = groupRepository;
+            CurrentUserContext = currentUserContext;
+        }
+        public bool IsValid(int id, Group group)
+        {
+            var result = true;
+            //If changing owner (which is an organization), validate the change
+            if (group.OwnerId != VALUE_NOT_SET)
```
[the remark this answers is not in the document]
davidmoore1: The group that comes into the form is the update request.  It only contains the fields that are changing.  For fields that are an int value, like the ids, they contain the value if they are changing or 0 if not.  This check is to see if the update request is changing the owner (organization) associated with the group.  If they aren't changing the owner, there is no need for a check.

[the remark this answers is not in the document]
NullVoxPopuli: might be good to notate that as it's not clear from usage, I think. idk

source/OptimaJet.DWKit.StarterApplication/Forms/Groups/UpdateForm.cs
```diff
@@ -0,0 +1,48 @@
...
+                var currentUser = UserRepository.GetByAuth0Id(CurrentUserContext.Auth0Id).Result;
+                var orgIds = currentUser.OrganizationIds ?? Enumerable.Empty<int>();
+                if (!orgIds.Contains(group.OwnerId))
+                {
+                    result = false;
+                    var original = GroupRepository.Get()
+                                                  .Where(g => g.Id == id)
+                                                  .Include(g => g.Owner)
+                                                  .FirstOrDefaultAsync().Result;
+                    var message = $"Group '{original.Name}': Current user not a member of group owner '{original.Owner.Name}'";
```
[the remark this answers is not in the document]
davidmoore1: OK.  Chris made the original error message and I just carried it along, but that works for me

source/OptimaJet.DWKit.StarterApplication/Forms/Groups/UpdateForm.cs
```diff
@@ -0,0 +1,48 @@
...
+                    AddError(message);
+                }
+            }
+
+            return result;
```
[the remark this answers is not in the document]
davidmoore1: Added base method.  Syntax is return Errors.Errors.Count == 0

source/OptimaJet.DWKit.StarterApplication/Forms/Projects/BaseProjectForm.cs
```diff
@@ -0,0 +1,50 @@
...
+            }
+            // The current user should be a member of the organization
+            var orgIds = CurrentUserOrgIds ?? Enumerable.Empty<int>();
+            if (!orgIds.Contains(Organization.Id))
+            {
+                result = false;
+                var message = $"Project '{ProjectName}': Current user not a member of project organization '{Organization.Name}'";
+                AddError(message);
+            }
+            return result;
```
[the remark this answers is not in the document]
davidmoore1: Changed as above

source/OptimaJet.DWKit.StarterApplication/Forms/Projects/UpdateForm.cs
```diff
@@ -0,0 +1,89 @@
...
+                // Fields that don't change, get their value from the original
+                if (project.OrganizationId != VALUE_NOT_SET)
+                {
+                    Organization = OrganizationRepository.Get()
+                            .Where(o => o.Id == project.OrganizationId)
+                            .FirstOrDefaultAsync().Result;
+                }
+                if (project.GroupId != VALUE_NOT_SET)
+                {
+                    Group = GroupRepository.Get()
```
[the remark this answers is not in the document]
davidmoore1: Same comment as before.  This is required because it is not present on an update

source/OptimaJet.DWKit.StarterApplication/Forms/Projects/UpdateForm.cs
```diff
@@ -0,0 +1,89 @@
...
+                            .Where(g => g.Id == project.GroupId)
+                           .Include(g => g.Owner).FirstOrDefaultAsync().Result;
+                }
+                if (project.OwnerId != VALUE_NOT_SET)
+                {
+                    Owner = UserRepository.Get()
```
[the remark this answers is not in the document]
davidmoore1: The owner is used in the BaseProjectForm.ValidateProject method (was isValid()).  Basically this comes down to "if the update changed the owner, get the new owner from the repository and perform the validation check with that."

source/SIL.AppBuilder.Portal.Backend.Tests/Acceptance/APIControllers/Groups/GetGroupsTest.cs
davidmoore1: The test Get method defaults the organization to an empty string if you don't specify an organization

NullVoxPopuli: right, but the comment is describing behavior that isn't written here. (there is no empty string present)

source/OptimaJet.DWKit.StarterApplication/Forms/Projects/UpdateForm.cs
```diff
@@ -0,0 +1,89 @@
...
+                                     .ThenInclude(g => g.Owner)
+                                  .FirstOrDefaultAsync().Result;
+                Owner = original.Owner;
+                Organization = original.Organization;
+                Group = original.Group;
+                // Set the fields to what would be the updated versions
+                // Fields that don't change, get their value from the original
+                if (project.OrganizationId != VALUE_NOT_SET)
+                {
+                    Organization = OrganizationRepository.Get()
```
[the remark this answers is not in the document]
davidmoore1: The Organization is being updated (!= VALUE_NOT_SET) therefore the organization to use to validate in the base project must come from the one specified by the update organization id rather than the one from the orginal record.

source/OptimaJet.DWKit.StarterApplication/Forms/Groups/UpdateForm.cs
```diff
@@ -0,0 +1,48 @@
...
+            CurrentUserContext = currentUserContext;
+        }
+        public bool IsValid(int id, Group group)
+        {
+            var result = true;
+            //If changing owner (which is an organization), validate the change
+            if (group.OwnerId != VALUE_NOT_SET)
+            {
+                var currentUser = UserRepository.GetByAuth0Id(CurrentUserContext.Auth0Id).Result;
+                var orgIds = currentUser.OrganizationIds ?? Enumerable.Empty<int>();
```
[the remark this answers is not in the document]
davidmoore1: Created IEnumerableExtension in Utility folder with this method.  Changed the forms and the repositories to call it.

source/OptimaJet.DWKit.StarterApplication/Services/GroupService.cs
```diff
@@ -0,0 +1,78 @@
...
+                {
+                    query = new JsonApiDotNetCore.Internal.Query.QuerySet();
+                    JsonApiContext.QuerySet = query;
+                }
+                var value = OrganizationContext.HasOrganization ? OrganizationContext.OrganizationId.ToString() : "";
+                query.Filters.Add(new JsonApiDotNetCore.Internal.Query.FilterQuery("organization-header", value, "="));
+                return await base.GetAsync();
+            }
+        }
+        public override async Task<Group> GetAsync(int id)
```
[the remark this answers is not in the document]
davidmoore1: No.  If this is referring to the GetAsync(id).  The change is having it go through Get(), which will call the function above which applies filters and does all the validation and then retrieves the record from the filtered list if it is there.  The default behavior does not apply filters and would not do the validation.  The base code calls GetAsync(id) on the repository and just gets the single record.

source/OptimaJet.DWKit.StarterApplication/Services/GroupService.cs
davidmoore1: Ditto

source/OptimaJet.DWKit.StarterApplication/Services/UserService.cs
davidmoore1: As with an earlier one, no.  The base calls the repository getAsync(id) directly without calling the filters etc.  The most straightforward way to implement the required checks was to getAsync() getting all and then pulling the requested one out of the list if it is still there.  Chris did send an email to you when we implemented asking if you had any objections.

source/OptimaJet.DWKit.StarterApplication/Services/UserService.cs
davidmoore1: Returning null at this point was resulting in a not found error.  I did go ahead and modify the code to send a json exception with a 404 code.  The unit tests did not require change.

source/SIL.AppBuilder.Portal.Backend.Tests/Acceptance/APIControllers/Groups/GetGroupTest.cs
davidmoore1: Added comment.  Attempting to get group that has an owner organization that the current user doesn't belong to.

source/SIL.AppBuilder.Portal.Backend.Tests/Acceptance/APIControllers/Groups/GetGroupTest.cs
davidmoore1: Organization header.  Changed name of test to reflect that.

source/SIL.AppBuilder.Portal.Backend.Tests/Acceptance/APIControllers/Groups/GetGroupsTest.cs
davidmoore1: Added comment.  Same as earlier.  The current user is not a member of the group owner organization.

source/SIL.AppBuilder.Portal.Backend.Tests/Acceptance/APIControllers/Groups/GetGroupsTest.cs
davidmoore1: Added comment.  Current user is not a member of the organization that is in the header for this test.

source/OptimaJet.DWKit.StarterApplication/Repositories/ProjectRepository.cs
```diff
@@ -16,67 +17,58 @@ public class ProjectRepository : ControllerRepository<Project>
...
         {
             this.OrganizationContext = organizationContext;
             this.CurrentUserContext = currentUserContext;
             this.UserRepository = userRepository;
+            this.GroupRepository = groupRepository;
+            this.OrganizationRepository = organizationRepository;
         }
-
-        public override IQueryable<Project> Get()
+        public override IQueryable<Project> Filter(IQueryable<Project> query, FilterQuery filterQuery)
```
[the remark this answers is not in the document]
davidmoore1: I worked with this for quite a while.  A few issues like the three GetWithOrganizationHeaders are not quite sufficiently alike to just have one.  The user one is somewhat different than the other two.  The final return filterQuery is also a problem since what it really returns at that point is a call to the base.Filter method which would be a bit hard to call from here I think.  You could maybe pass another Func that would point.  Some other things.  Finally decided after working on it for quite a while that I needed to get the other issues resolved and this one really is not creating any performance issues as is.

source/SIL.AppBuilder.Portal.Backend.Tests/Acceptance/APIControllers/Groups/GetGroupsTest.cs
davidmoore1: Comment updated

source/OptimaJet.DWKit.StarterApplication/Services/GroupService.cs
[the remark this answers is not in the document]
NullVoxPopuli: I See, https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/master/src/JsonApiDotNetCore/Services/EntityResourceService.cs#L121

cool.

source/OptimaJet.DWKit.StarterApplication/Services/GroupService.cs
```diff
@@ -0,0 +1,80 @@
...
+        {
+            var groups = await GetAsync();
+            return groups.SingleOrDefault(g => g.Id == id);
+        }
+        public override async Task<Group> UpdateAsync(int id, Group resource)
+        {
+            var updateForm = new Forms.Groups.UpdateForm(UserRepository,
```
NullVoxPopuli: is there a particular reason for using the (nearly) fully qualified name?

source/OptimaJet.DWKit.StarterApplication/Services/GroupService.cs
```diff
@@ -0,0 +1,80 @@
...
+                                                         GroupRepository, 
+                                                         CurrentUserContext);
+            if (!updateForm.IsValid(id, resource))
+            {
+                throw new JsonApiException(updateForm.Errors);
+            }
+            // Make sure that group may be accessed by current user
+            var group = await GetAsync(id);
```
NullVoxPopuli: what does this comment have to do with the code immediately below it? The current user is not involved here.

source/OptimaJet.DWKit.StarterApplication/Services/HttpOrganizationContext.cs
```diff
@@ -11,7 +11,7 @@ public class HttpOrganizationContext : IOrganizationContext
         public HttpContext HttpContext { get; set; }
         public OrganizationService OrganizationService { get; set; }
 
-        private bool parsed;
+        private bool parsed = false;
```
NullVoxPopuli: why was this change made?

source/OptimaJet.DWKit.StarterApplication/Services/ProjectService.cs
```diff
@@ -0,0 +1,102 @@
...
+            if (OrganizationContext.SpecifiedOrganizationDoesNotExist)
+            {
+                return Enumerable.Empty<Project>().AsQueryable();
+            }
+            else
+            {
+                var query = JsonApiContext.QuerySet;
+                if (query == null)
+                {
+                    query = new JsonApiDotNetCore.Internal.Query.QuerySet();
```
NullVoxPopuli: any reason for the fully qualified namespace?

source/OptimaJet.DWKit.StarterApplication/Services/ProjectService.cs
```diff
@@ -0,0 +1,102 @@
...
+                    JsonApiContext.QuerySet = query;
+                }
+                var value = OrganizationContext.HasOrganization ? OrganizationContext.OrganizationId.ToString() : "";
```
NullVoxPopuli: I've seen this in a few places. I wonder if there should be a method on the OrganizationContext.

source/OptimaJet.DWKit.StarterApplication/Services/ProjectService.cs
```diff
@@ -0,0 +1,102 @@
...
+                query.Filters.Add(new JsonApiDotNetCore.Internal.Query.FilterQuery("organization-header", value, "="));
+                return await base.GetAsync();
```
NullVoxPopuli: I think the body of this whole method could be made generic, and maybe be called:

```
GetScopedToOrganization<T>();
```

or something like that, idk.

source/OptimaJet.DWKit.StarterApplication/Services/ProjectService.cs
```diff
@@ -0,0 +1,102 @@
...
+                                                           GroupRepository,
+                                                           CurrentUserContext,
+                                                           OrganizationRepository,
+                                                           ProjectRepository);
+            if (!updateForm.IsValid(id, resource))
+            {
+                throw new JsonApiException(updateForm.Errors);
+            }
+            // Make sure that group may be accessed by current user
+            var project = await GetAsync(id);
```
NullVoxPopuli: why get the project again? the UpdateForm already needs it yeah?

source/OptimaJet.DWKit.StarterApplication/Services/ProjectService.cs
```diff
@@ -0,0 +1,102 @@
...
+            if (project == null)
+            {
+                throw new JsonApiException(404, $"Project Id '{id}' not found."); ;
+            }
+            return await base.UpdateAsync(id, resource);
```
NullVoxPopuli: I wonder if it'd make more sense to rename this `resource` argument to be more contextual to it's use / origin.

maybe: `ModelFromParams` or something.

source/OptimaJet.DWKit.StarterApplication/Services/UserService.cs
NullVoxPopuli: Not found is a correct error. -- what was wrong without the null? like, what was the response body content?

source/OptimaJet.DWKit.StarterApplication/Services/UserService.cs
```diff
@@ -11,14 +15,60 @@ namespace OptimaJet.DWKit.StarterApplication.Services
...
         {
             this.EntityRepository = (UserRepository)entityRepository;
+            JsonApiContext = jsonApiContext;
+            OrganizationContext = organizationContext;
+            CurrentUserContext = currentUserContext;
+        }
+
+        public override async Task<IEnumerable<User>> GetAsync()
+        {
+            if (OrganizationContext.SpecifiedOrganizationDoesNotExist)
```
NullVoxPopuli: yeah, this could totes be extracted :)

source/OptimaJet.DWKit.StarterApplication/Utility/IEnumerableExtensions.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace OptimaJet.DWKit.StarterApplication.Utility
+{
+    public static class IEnumerableExtensions
+    {
+        public static IEnumerable<int> OrEmpty(this IEnumerable<int> list)
```
NullVoxPopuli: can you change `int` to be generic. :)

source/OptimaJet.DWKit.StarterApplication/Forms/Projects/CreateForm.cs
```diff
@@ -0,0 +1,60 @@
...
+                       .Where(g => g.Id == project.GroupId)
+                       .Include(g => g.Owner).FirstOrDefaultAsync().Result;
+                Owner = UserRepository.Get()
+                        .Where(u => u.Id == project.OwnerId)
+                        .Include(u => u.GroupMemberships)
+                            .ThenInclude(gm => gm.Group)
+                        .Include(u => u.OrganizationMemberships)
+                            .ThenInclude(om => om.Organization)
+                        .FirstOrDefaultAsync().Result;
+                ProjectName = project.Name;
```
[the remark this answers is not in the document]
davidmoore1: Not any more.  Got eliminated when we simplified the messages last time and I didn't notice.

source/OptimaJet.DWKit.StarterApplication/Forms/Groups/UpdateForm.cs
```diff
@@ -0,0 +1,47 @@
...
+        public bool IsValid(int id, Group group)
+        {
+            //If changing owner (which is an organization), validate the change
+            if (group.OwnerId != VALUE_NOT_SET)
+            {
+                var currentUser = UserRepository.GetByAuth0Id(CurrentUserContext.Auth0Id).Result;
+                var orgIds = currentUser.OrganizationIds.OrEmpty();
+                if (!orgIds.Contains(group.OwnerId))
+                {
+                    var original = GroupRepository.Get()
```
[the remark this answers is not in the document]
davidmoore1: Yeah, another case of the simplified message change and not noticing that the variable wasn't needed.

source/OptimaJet.DWKit.StarterApplication/Forms/Projects/CreateForm.cs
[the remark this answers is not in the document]
NullVoxPopuli: well, should there be a requirement for project name to be present? like as a validation?

`RequireName(project)`?

source/OptimaJet.DWKit.StarterApplication/Forms/Projects/BaseProjectForm.cs
```diff
@@ -0,0 +1,45 @@
...
+
+        protected void ValidateProject()
+        {
+            if ((Organization == null) || (Group == null) || (Owner == null))
+            {
+                // Allowing it to return in these cases should cause the base to hit normal
+                // foreign key failures
+                return;
+            }
+            if (Organization != Group.Owner)
```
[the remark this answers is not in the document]
davidmoore1: Changed the name to project owner

source/OptimaJet.DWKit.StarterApplication/Forms/Projects/UpdateForm.cs
```diff
@@ -0,0 +1,83 @@
...
+                if (project.OwnerId != VALUE_NOT_SET)
+                {
+                    Owner = UserRepository.Get()
+                            .Where(u => u.Id == project.OwnerId)
+                            .Include(u => u.OrganizationMemberships)
+                                .ThenInclude(om => om.Organization)
+                            .FirstOrDefaultAsync().Result;
+                }
+
+                ProjectName = project.Name;
```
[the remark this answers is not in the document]
davidmoore1: Another instance of the simplified messages.  Removed variable from base and this refernce.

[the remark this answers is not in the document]
NullVoxPopuli: but maybe it should be used? like.. a project needs a name, yeah?

source/OptimaJet.DWKit.StarterApplication/Forms/Projects/UpdateForm.cs
```diff
@@ -0,0 +1,83 @@
...
+                var currentUser = UserRepository.GetByAuth0Id(CurrentUserContext.Auth0Id).Result;
```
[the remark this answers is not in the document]
davidmoore1: Added it to the base form constructor with supporting changes

source/OptimaJet.DWKit.StarterApplication/Utility/IEnumerableExtensions.cs
davidmoore1: Done

source/OptimaJet.DWKit.StarterApplication/Services/ProjectService.cs
davidmoore1: Done

source/OptimaJet.DWKit.StarterApplication/Services/ProjectService.cs
davidmoore1: we were etting the project with the ID calls GetAsync which runs the filter through the service to ensures that the current user is allowed to access the record.  I changed this for the updates to add another validation routine to the base form and avoiding duplicate queries as much as possible

source/OptimaJet.DWKit.StarterApplication/Services/UserService.cs
davidmoore1: Done

source/OptimaJet.DWKit.StarterApplication/Services/ProjectService.cs
davidmoore1: With the GetAsync method extraction, this only appears in one place now

source/OptimaJet.DWKit.StarterApplication/Services/ProjectService.cs
davidmoore1: Don't know at this point.  Removed it.

source/OptimaJet.DWKit.StarterApplication/Services/HttpOrganizationContext.cs
davidmoore1: It was changed as part of a debug session and accidentally checked in.  I've put it back.  Was really weird.  I had put a breakpoint in Parse, and it always came up true and would never execute the method.  The only way to make it work was to remove the breakpoint.  Anyway, didn't mean that to get checked in.

source/OptimaJet.DWKit.StarterApplication/Services/GroupService.cs
davidmoore1: This is another case where a get was being performed so that the filters would run and make sure that the current user was a member or the group owner id and that the group owner matched the organization header  entry if it was set to a value.  For updates, that has now been moved to the base validation form.

source/OptimaJet.DWKit.StarterApplication/Services/GroupService.cs
davidmoore1: This is because until yesterday this was in the BackendServicesExtension and forms with the same name were being created for Projects and Groups in the same routine.  When I cut them out of there and changed it to a get new, I didn't think about the fact that there was no longer a need for the longer signature. Fixed

source/OptimaJet.DWKit.StarterApplication/Repositories/ProjectRepository.cs
```diff
@@ -16,67 +17,58 @@ public class ProjectRepository : ControllerRepository<Project>
...
         }
 
-        public override async Task<Project> CreateAsync(Project entity)
-        {
-            ValidateOrganization(entity);
-            return await base.CreateAsync(entity);
-        }
 
-        private void ValidateOrganization(Project project)
+        private IQueryable<Project> GetWithOrganizationHeader(IQueryable<Project> query, string value)
```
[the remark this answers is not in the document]
davidmoore1: Not identical.  The queries in the methods are based on different fields

source/OptimaJet.DWKit.StarterApplication/Services/HttpOrganizationContext.cs
NullVoxPopuli: sounds like some race condition could be present or something

source/OptimaJet.DWKit.StarterApplication/Services/GroupService.cs
NullVoxPopuli: ok cool. 

It's just a comment should be directly making things _more_ clear, or the comment isn't needed / worth having :)

chrisvire:
@NullVoxPopuli, what is left on this PR?

source/OptimaJet.DWKit.StarterApplication/Services/ProjectService.cs
```diff
@@ -0,0 +1,85 @@
...
+        {
+            var createForm = new CreateForm(UserRepository,
+                                           GroupRepository,
+                                           CurrentUserContext,
+                                           OrganizationRepository);
+            if (!createForm.IsValid(resource))
+            {
+                throw new JsonApiException(createForm.Errors);
+            }
+            var retval = await base.CreateAsync(resource);
```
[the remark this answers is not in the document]
davidmoore1: Probably part of a debug session that didn't get cleaned up where I was putting in a breakpoint to review the value.

source/OptimaJet.DWKit.StarterApplication/Services/ProjectService.cs
```diff
@@ -0,0 +1,85 @@
...
+                                           GroupRepository,
+                                           CurrentUserContext,
+                                           OrganizationRepository,
+                                           OrganizationContext,
+                                           ProjectRepository);
+            if (!updateForm.IsValid(id, resource))
+            {
+                throw new JsonApiException(updateForm.Errors);
+            }
+             return await base.UpdateAsync(id, resource);
```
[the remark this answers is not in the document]
davidmoore1: I think I just left it at resource to avoid getting a warning telling me that the method it is overriding calls it something else

PR #109, opened 2018-08-28
Filtering user table 

giancorzo:
https://developertown.visualstudio.com/SIL%20International%20Scriptoria/_workitems/edit/28331

- [x] Add filter based on selected Organization
- [x] Add tests

### DIFF
--- a/source/SIL.AppBuilder.Portal.Frontend/src/ui/components/user-table/index.tsx
+++ b/source/SIL.AppBuilder.Portal.Frontend/src/ui/components/user-table/index.tsx
@@ -8,9 +8,6 @@ import { compose } from 'recompose';
 import Table from './table';
 import { withData } from './data';
 
-  export default compose(
-//    withSorting({ defaultSort: 'name' }),
-//    withPagination,
-//    withFiltering,
-    withData
-  )(Table);
+export default compose(
+  withData
+)(Table);
--- a/source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/users/list/__tests__/filter-by-organization-acceptance-test.tsx
+++ b/source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/users/list/__tests__/filter-by-organization-acceptance-test.tsx
@@ -1,0 +1,140 @@
+
+import { describe, it, beforeEach } from '@bigtest/mocha';
+import { visit, location } from '@bigtest/react';
+import { expect } from 'chai';
+
+import { setupApplicationTest, setupRequestInterceptor, useFakeAuthentication } from 'tests/helpers';
+
+import app from 'tests/helpers/pages/app';
+import switcher from '@ui/components/sidebar/org-switcher/__tests__/page';
+import page from './page';
+
+describe('Acceptance | User list | Filtering users by organization', () => {
+
+  setupApplicationTest();
+  setupRequestInterceptor();
+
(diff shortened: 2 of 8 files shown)


source/SIL.AppBuilder.Portal.Frontend/src/ui/components/user-table/index.tsx
```diff
@@ -8,9 +8,6 @@ import { compose } from 'recompose';
 import Table from './table';
 import { withData } from './data';
 
-  export default compose(
-//    withSorting({ defaultSort: 'name' }),
-//    withPagination,
```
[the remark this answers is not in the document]
giancorzo: Yes, eventually

source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/users/list/__tests__/filter-by-organization-acceptance-test.tsx
```diff
@@ -0,0 +1,140 @@
...
+
+        expect(page.usernames().length).to.equal(3);
+
+        const usernames = page.usernames();
+        const text = usernames.map(u => u.text).join();
+
+        expect(text).to.include('fake fake');
+        expect(text).to.include('One fake');
+        expect(text).to.include('Two fake');
+      });
```
NullVoxPopuli: there should be a nested test in here for switching the organization after already viewing the users screen, and making sure users stop showing, because that was the bug :)

NullVoxPopuli: The current set of tests only test what happens if you start off by visiting the users page with only a particular org selected. Gotta test the switch :)

giancorzo: Isnt the next test testing that? The first one test that you see the page then i change the organization and I get New result. Perhaps a test that test the negative path that if I change the organization it wont add up?

NullVoxPopuli: according to the (lack of) nesting of the describe blocks, they are separate scenarios and are independent of each other.

This is the flow that needs to exist as a test:
 - Select All Organizations
    - View Users Page
       -> should see all users
    - Select a specific Organization (still on the users page)
       -> should see a subset of the previously asserted users

source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/users/list/__tests__/filter-by-organization-acceptance-test.tsx
```diff
@@ -0,0 +1,134 @@
...
+          });
+
+          it('Only the user that belongs to the same organization are displayed', () => {
+
+            expect(page.usernames().length).to.equal(2);
+
+            const usernames = page.usernames();
+            const text = usernames.map(u => u.text).join();
+
+            expect(text).to.include('fake fake');
```
NullVoxPopuli: there should be additional assertions here telling me that the other two users aren't here, cause as-is, I don't have enough information to make that assumption.

PR #116, opened 2018-09-07
Project toggle settings (automaticRebuild, allowDownload)

giancorzo:


source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/project/settings/with-settings.tsx
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    toggleField = async (projectId, fieldName, newToggleState) => {
+
+      const { updateStore } = this.props;
+
+      try {
+
+        updateStore(t => t.replaceAttribute(
+          { type: PROJECT, id: projectId },
```
NullVoxPopuli: you'll want to use the actual `project` from `props` here instead of this type/id object.

NullVoxPopuli: see: data/containers/project/with-data-actions#updateAttribute

PR #118, opened 2018-09-07
Add Product API

davidmoore1:


source/OptimaJet.DWKit.StarterApplication/Forms/Products/BaseProductForm.cs
```diff
@@ -0,0 +1,45 @@
...
+            if ((project.Organization == null))
+            {
+                var message = $"The project for this product does not have a valid organization";
+                AddError(message);
+                return; // Can't proceed further without causing exceptions
+            }
+
+            if (!project.Organization.ProductDefinitionIds.Contains(productDefinition.Id))
+            {
+                var message = $"This product is not defined for the product's project organization";
```
chrisvire: I don't understand this error message.  What this check seems to do is very that ProductDefinition is permitted for the Organization.

davidmoore1: No product definition I translated as the product is not defined for.  These are all one step removed relationships which made it difficult to figure out how to put this.  Would "The product definition for this product is not permitted for the product's project organization?" be acceptable?

chrisvire: How about: "This product is not permitted for this organization."

davidmoore1: How about "This product is not permitted for the product's project organization"

source/OptimaJet.DWKit.StarterApplication/Forms/Products/BaseProductForm.cs
```diff
@@ -0,0 +1,45 @@
...
+        }
+        protected void ValidateProduct(Project project, ProductDefinition productDefinition)
+        {
+            if ((project == null) || (productDefinition == null))
+            {
+                // Allowing it to return in these cases should cause the base to hit normal
+                // foreign key failures
+                return;
+            }
+            if ((project.Organization == null))
```
NullVoxPopuli: since this is a product form, should we assume that a project has already been validated / has already been persisted to the database? if so, we may be able to skip this check.

source/OptimaJet.DWKit.StarterApplication/Forms/Products/BaseProductForm.cs
```diff
@@ -0,0 +1,45 @@
...
+
+            if (!project.Organization.ProductDefinitionIds.Contains(productDefinition.Id))
+            {
+                var message = $"This product is not permitted for the organization";
+                AddError(message);
+            }
+            // The current user should be a member of the organization
+            if (!CurrentUserOrgIds.Contains(project.Organization.Id))
+            {
+                var message = ("The current user is not a member of the product organization");
```
NullVoxPopuli: message should say project?

source/OptimaJet.DWKit.StarterApplication/Utility/RepositoryExtensions.cs
```diff
@@ -27,8 +27,7 @@ public static class RepositoryExtensions
 
             if (isTargetParam)
             {
-                var result = getMethod(query, value, userRepository, currentUserContext, query1, query2);//filterOn(query, value);
-                return result;
+                return getMethod(query, value, userRepository, currentUserContext, query1, query2);//filterOn(query, value);
```
NullVoxPopuli: while we're here, query1 and query2 could really use some better names. :-\

source/OptimaJet.DWKit.StarterApplication/Forms/Products/BaseProductForm.cs
davidmoore1: Removed the check

source/OptimaJet.DWKit.StarterApplication/Forms/Products/BaseProductForm.cs
davidmoore1: From what Chris was after me about in the other message, I guess this one should just say that the current user is not a member of the organization. The reasons for removing project in the other message was "It just happens to be connected to a project.  It wouldn’t be allowed for any project of the organization so “project” is just a container of products and not relevant to the check. The project is inherently tied to a organization.  It is a container.  It really isn’t relevant to the check other than the way to navigate to the organization.
You are really checking, “Does this organization allow this type of product”.  I would assume that the same rationale would apply here and change the message to "The current user is not a member of the organization"

PR #120, opened 2018-09-07
add some better debuggability

NullVoxPopuli:
image


source/SIL.AppBuilder.Portal.Frontend/src/data/containers/resources/project/with-data-actions.tsx
```diff
@@ -81,5 +83,8 @@ export function withDataActions<T>(WrappedComponent) {
     }
   }
 
-  return withOrbit({})(ProjectDataActionWrapper);
+  return compose(
+    withOrbit({}),
+    requireProps('project')
```
NullVoxPopuli: since this is specific to props, I could bring myself to call it just `assert`

PR #126, opened 2018-09-12
Prepare for BuildEngine API

davidmoore1:
Update APK Release
Wrap Hangfire for testing

### DIFF
--- a/source/OptimaJet.DWKit.StarterApplication/Services/Contracts/IHangfireWrapper.cs
+++ b/source/OptimaJet.DWKit.StarterApplication/Services/Contracts/IHangfireWrapper.cs
@@ -1,0 +1,10 @@
+﻿using System;
+using Hangfire;
+
+namespace OptimaJet.DWKit.StarterApplication.Services
+{
+    public interface IHangfireWrapper
+    {
+        IBackgroundJobClient BackgroundJobClient { get; }
+    }
+}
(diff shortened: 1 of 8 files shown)


source/OptimaJet.DWKit.StarterApplication/Services/Contracts/IHangfireWrapper.cs
```diff
@@ -0,0 +1,10 @@
+﻿using System;
+using Hangfire;
+
+namespace OptimaJet.DWKit.StarterApplication.Services
+{
+    public interface IHangfireWrapper
+    {
+        IBackgroundJobClient BackgroundJobClient { get; }
```
chrisvire: We will also need RecurringJob.

davidmoore1: Yes, this gives the framework for what I need for the first step.  You are right that something along that line will need to be added when we add the recurring job

source/OptimaJet.DWKit.StarterApplication/Services/Contracts/IHangfireWrapper.cs
```diff
@@ -0,0 +1,10 @@
+﻿using System;
+using Hangfire;
+
+namespace OptimaJet.DWKit.StarterApplication.Services
+{
+    public interface IHangfireWrapper
```
chrisvire: I would be good to have comments on why using IHangfireWrapper instead of injecting IBackgroundJobClient.

http://www.jondjones.com/c-sharp-bootcamp/tdd/misc-unit-testing-snippets/how-to-unit-test-your-hangfire-tasks

PR #127, opened 2018-09-12
convert translations to json, and fix a bug with updating the set locale

NullVoxPopuli:
/cc @chrisvire


source/SIL.AppBuilder.Portal.Frontend/src/translations/index.tsx
```diff
@@ -1,12 +1,11 @@
-import * as React from 'react';
 import * as i18n from 'i18next';
 import * as ICU from 'i18next-icu';
-import * as Backend from 'i18next-xhr-backend';
+// import * as Backend from 'i18next-xhr-backend';
```
giancorzo: we should remove it this type of comments unless we are going to revert it later

NullVoxPopuli: yeah, backend loaded translations are a possibility

PR #129, opened 2018-09-12
WIP: [Feature]: Project Directory Filtering and Search

NullVoxPopuli:
### [26305 - View Project Directory](https://developertown.visualstudio.com/SIL%20International%20Scriptoria/_workitems/edit/26305)
 - [ ] ( FILTER ) All projects for the organization are listed
    - [ ] Test case: make sure filter is working for a single org
 - [ ] ( SORTING ) User can sort projects 
 - [ ] ( DISPLAY ) Product icon types visible 
    - **Q: do products coming from the database have a type?**
       - _A: relation is: project.products[].product-definition.application-type_
 - [x] ( DISPLAY) Number of projects in directory displayed
 - [x] ( DISPLAY ) List should be paginated, displaying 20 projects per page

### [26411 - Search Help](https://developertown.visualstudio.com/SIL%20International%20Scriptoria/_workitems/edit/26411)
- [ ] ( FILTER ) Search assistant should display what search options are 

### [26422 - Filter Project Directory](https://developertown.visualstudio.com/SIL%20International%20Scriptoria/_workitems/edit/26422)
 - [ ] ( FILTER ) User can choose range of results displayed, and list of projects adjust based on dates selected 
   - [ ] User cannot select a 'to' date in the future
   - [x] User cannot select a 'from' date past today
 - [ ] (FILTER ) When projects are filtered by a product type, only projects containing that product type are displayed 
 - [ ] ( FILTER ) When projects are filtered by a org, only projects inside of that org are displayed 
    - [ ] Should be allowed even when the current user is not a member of the organization
 - [ ] ( FILTER ) When filter has been set for a certain language, only projects in that language appear 

These 3 stories are combined into one PR because they are all every tightly coupled to each other. As I work through this and whatever issues come up, I may split out another PR.

## Dependent On:

Note: when querying projects, the backend must support the following (outside the scope of this PR):
 - [ ] Query projects outside of the users organization
 - [ ] Prevent projects from showing that are private


NullVoxPopuli:
Problems Encountered:
1. Updated from master 2018-09-20
   - included resources no longer are returned from the users controllers' current-user endpoint.

NullVoxPopuli:
image

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/layout/index.tsx
```diff
@@ -37,6 +38,11 @@ class Layout extends React.Component<IOwnProps> {
 
     if (window.innerWidth > SM_BREAKPOINT && !isSidebarVisible) {
       showSidebar();
+    } else {
+      // intentionally do not save this state (by not using redux)
+      // this is more for dev purposes where we are constantly 
+      // resizing the browser
+      hideSidebar();
```
giancorzo: why we are doing this?

NullVoxPopuli: because every time I resized the window down to a smaller size, the sidebar would be in my way, when it should be hidden by default.

NullVoxPopuli:
Tests failing only on Travis. :-\

PR #138, opened 2018-09-17
Project table column customization

giancorzo:


source/SIL.AppBuilder.Portal.Frontend/src/ui/components/project-table/withTableColumns.tsx
```diff
@@ -0,0 +1,111 @@
...
+    return {
+      setColumnSelection: (column) => dispatch(setColumnSelectionData(tableName, column))
+    };
+  }
+
+  return InnerComponent => {
+
+    class WrapperComponent extends React.Component<IProvidedProps & IDataProps> {
+
+      updateColumnSelection = (column) => {
```
dalinarkholin: nice! 🎉

dalinarkholin: So we are in the context of a component here. Possible to extract this logic into your helper lib? Same with `isInSelectedColumns`. Probably a few other things, just to try and keep the components simple.

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/project-table/withTableColumns.tsx
```diff
@@ -0,0 +1,111 @@
...
+      columnWidth = (type = 'header') => {
+        const { selectedColumns } = this.props;
+        const columnCount = selectedColumns.filter(c => c.type === type);
+        return (100 / ( columnCount.length + 1) ); // +1 because it includes project column
+      }
+
+      render() {
+
+        const columns = [
+          { id: 'owner', type: 'header' },
```
dalinarkholin: I have see columns defined in a couple places. Possible to extract definition into a enum? (Could be a later improvement)

giancorzo: Yeah, I could move it to redux-store but my guess is that this info will eventually get into the database as a user preference.

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/project-table/withTableColumns.tsx
giancorzo: Actually, I moved this code from the visual component into this HOC. I could extract it to a helper function but I'm not going to reuse it anywhere else so I think for now I leave it here where it adds context.

PR #149, opened 2018-09-23
Fix logout interface

giancorzo:


source/SIL.AppBuilder.Portal.Frontend/src/data/index.ts
```diff
@@ -24,7 +24,7 @@ export {
 } from './helpers';
 
 export { withLoader } from './containers/with-loader';
-export { withLogout, IProvidedProps } from '@data/containers/with-logout';
```
NullVoxPopuli: is this used anywhere?
I'm also somewhat against this change, because - within the context of a component, we know what the purpose of that interface is -- the props that are provided. :)

giancorzo: Yes it's used in the header component (user-dropdown.tsx)

Are you against the name or to export the interface? I don't main revert it to IProvidedProps I did this PR because in user-dropdown I'm using ILogoutProps so this was mark as an error.

PR #153, opened 2018-09-25
Monitor connections for availability

davidmoore1:


source/OptimaJet.DWKit.StarterApplication/BackendServicesExtension.cs
```diff
@@ -17,6 +17,7 @@
 using OptimaJet.DWKit.StarterApplication.Models;
 using OptimaJet.DWKit.StarterApplication.Repositories;
 using OptimaJet.DWKit.StarterApplication.Services;
+using SIL.AppBuilder.BuildEngineApiClient;
```
chrisvire: Please remove this "using" and use the fully qualified name so we don't have "Model.Project"

PR #165, opened 2018-09-28
Fix tests for ApplicationType changes

davidmoore1:


NullVoxPopuli:
Why does this fix things? What wasn't working before?

chrisvire:
@NullVoxPopuli, `String type` changed to `int type-id` or specifying the relationship object.  

@davidmoore1, did you try changing `type = "scriptureappbuilder"` to `type-id = 1` ?

NullVoxPopuli:
@chrisvire thanks!
just seemed strange that this is _mostly_ a conversion from anonymous objects to Dictionaries and Lists

chrisvire:
@NullVoxPopuli I think he had to change to Dictionaries due to not being able to specify "type" as a property in an anonymous object (at least that is what I am supposing).

chrisvire:
@NullVoxPopuli, per json-api, would it be better to assign "Type" via the `type-id` or the `type` relationship (or for testing purposes, both should be done)?

davidmoore1:
The original reason for making the changes was that I was having some problems creating the record via ARC and realized the variable had changed.  I went to the Create tests to see what the correct syntax was and it was clearly wrong.  The tests worked because there were no asserts for the value of type or type-id, so that was added.  I'm surprised they worked trying to set type with a string value.  type-id is set by setting the relationship, which is what the assert verifies.

PR #155, opened 2018-09-25
Actual work on Filtering on the Project Directory -- finally

NullVoxPopuli:
Description pulled from https://github.com/sillsdev/appbuilder-portal/pull/129
And supercedes: https://github.com/sillsdev/appbuilder-portal/pull/150
And supercedes: https://github.com/sillsdev/appbuilder-portal/pull/154

---------------------------------------------------

### [26305 - View Project Directory](https://developertown.visualstudio.com/SIL%20International%20Scriptoria/_workitems/edit/26305)
 - [x] ( FILTER ) All projects for the organization are listed
    - [x] Test case: make sure filter is working for a single org
 - [ ] ( SORTING ) User can sort projects 
 - [ ] ( DISPLAY ) Product icon types visible 
    - **Q: do products coming from the database have a type?**
       - _A: relation is: project.products[].product-definition.application-type_
 - [x] ( DISPLAY) Number of projects in directory displayed
 - [x] ( DISPLAY ) List should be paginated, displaying 20 projects per page

### [26411 - Search Help](https://developertown.visualstudio.com/SIL%20International%20Scriptoria/_workitems/edit/26411)
- [ ] ( FILTER ) Search assistant should display what search options are 

### [26422 - Filter Project Directory](https://developertown.visualstudio.com/SIL%20International%20Scriptoria/_workitems/edit/26422)
 - [ ] ( FILTER ) User can choose range of results displayed, and list of projects adjust based on dates selected 
   - [ ] User cannot select a 'to' date in the future
   - [x] User cannot select a 'from' date past today
 - [ ] (FILTER ) When projects are filtered by a product type, only projects containing that product type are displayed 
 - [x] ( FILTER ) When projects are filtered by a org, only projects inside of that org are displayed 
    - [x] Should be allowed even when the current user is not a member of the organization
 - [ ] ( FILTER ) When filter has been set for a certain language, only projects in that language appear 

These 3 stories are combined into one PR because they are all every tightly coupled to each other. As I work through this and whatever issues come up, I may split out another PR.

## Dependent On:

Note: when querying projects, the backend must support the following (outside the scope of this PR):
 - [x] Query projects outside of the users organization
 - [x] Prevent projects from showing that are private

## Unexpected things:

2018-09-25
 - Can't update database with `./run bootstrap` https://github.com/aspnet/EntityFrameworkCore/issues/12220

20180-09-28
 - Inconsistencies with tests vs reality. 
   - We should probably be testing that request headers are set.

PR #150, opened 2018-09-24
Filtering Part 2: Backend support for project directory.

NullVoxPopuli:
Description pulled from https://github.com/sillsdev/appbuilder-portal/pull/129

---------------------------------------------------

### [26305 - View Project Directory](https://developertown.visualstudio.com/SIL%20International%20Scriptoria/_workitems/edit/26305)
 - [x] ( FILTER ) All projects for the organization are listed
    - [ ] Test case: make sure filter is working for a single org
 - [ ] ( SORTING ) User can sort projects 
 - [ ] ( DISPLAY ) Product icon types visible 
    - **Q: do products coming from the database have a type?**
       - _A: relation is: project.products[].product-definition.application-type_
 - [x] ( DISPLAY) Number of projects in directory displayed
 - [x] ( DISPLAY ) List should be paginated, displaying 20 projects per page

### [26411 - Search Help](https://developertown.visualstudio.com/SIL%20International%20Scriptoria/_workitems/edit/26411)
- [ ] ( FILTER ) Search assistant should display what search options are 

### [26422 - Filter Project Directory](https://developertown.visualstudio.com/SIL%20International%20Scriptoria/_workitems/edit/26422)
 - [ ] ( FILTER ) User can choose range of results displayed, and list of projects adjust based on dates selected 
   - [ ] User cannot select a 'to' date in the future
   - [x] User cannot select a 'from' date past today
 - [ ] (FILTER ) When projects are filtered by a product type, only projects containing that product type are displayed 
 - [ ] ( FILTER ) When projects are filtered by a org, only projects inside of that org are displayed 
    - [ ] Should be allowed even when the current user is not a member of the organization
 - [ ] ( FILTER ) When filter has been set for a certain language, only projects in that language appear 

These 3 stories are combined into one PR because they are all every tightly coupled to each other. As I work through this and whatever issues come up, I may split out another PR.

## Dependent On:

Note: when querying projects, the backend must support the following (outside the scope of this PR):
 - [x] Query projects outside of the users organization
 - [x] Prevent projects from showing that are private

PR #167, opened 2018-09-28
Backend Filtering Support

NullVoxPopuli:
Description pulled from https://github.com/sillsdev/appbuilder-portal/pull/129
And supercedes: https://github.com/sillsdev/appbuilder-portal/pull/150
And supercedes: https://github.com/sillsdev/appbuilder-portal/pull/154
And supercedes: https://github.com/sillsdev/appbuilder-portal/pull/155

---------------------------------------------------

### [26305 - View Project Directory](https://developertown.visualstudio.com/SIL%20International%20Scriptoria/_workitems/edit/26305)
 - [x] ( FILTER ) All projects for the organization are listed
    - [x] Test case: make sure filter is working for a single org
 - [ ] ( SORTING ) User can sort projects 
 - [ ] ( DISPLAY ) Product icon types visible 
    - **Q: do products coming from the database have a type?**
       - _A: relation is: project.products[].product-definition.application-type_
 - [x] ( DISPLAY) Number of projects in directory displayed
 - [x] ( DISPLAY ) List should be paginated, displaying 20 projects per page

### [26411 - Search Help](https://developertown.visualstudio.com/SIL%20International%20Scriptoria/_workitems/edit/26411)
- [x] ( FILTER ) Search assistant should display what search options are 

### [26422 - Filter Project Directory](https://developertown.visualstudio.com/SIL%20International%20Scriptoria/_workitems/edit/26422)
 - [x] ( FILTER ) User can choose range of results displayed, and list of projects adjust based on dates selected 
   - [x] User cannot select a 'to' date in the future
   - [x] User cannot select a 'from' date past today
 - [ ] (FILTER ) When projects are filtered by a product type, only projects containing that product type are displayed 
 - [x] ( FILTER ) When projects are filtered by a org, only projects inside of that org are displayed 
    - [x] Should be allowed even when the current user is not a member of the organization
 - [x] ( FILTER ) When filter has been set for a certain language, only projects in that language appear 

These 3 stories are combined into one PR because they are all every tightly coupled to each other. As I work through this and whatever issues come up, I may split out another PR.

## Dependent On:

Note: when querying projects, the backend must support the following (outside the scope of this PR):
 - [x] Query projects outside of the users organization
 - [x] Prevent projects from showing that are private

## Unexpected things:

2018-09-28
 - jsonapi-dotnet-core doesn't have the filter apis implemented that allow easy filtering on relationships
   - see: https://github.com/json-api-dotnet/JsonApiDotNetCore/issues/238
   - this will greatly increase the time of implementing filtering on product attributes. :-\

2018-10-1
 - need to implement query/filter handlers
 - a test is failing only in docker, but not on local dotnet


NullVoxPopuli:
currently, 121 tests.

source/SIL.AppBuilder.Portal.Frontend/package.json
```diff
@@ -55,13 +56,13 @@
     "react-emotion": "^9.2.6",
     "react-i18next": "7.7.0",
     "react-notify-toast": "^0.5.0",
-    "react-orbitjs": "github:developertown/react-orbitjs#f28784f4ca7fcb53540692c2e52496448428beb9",
+    "react-orbitjs": "github:NullVoxPopuli/react-orbitjs#50e67610bd8748bddeb7d1e0de1615a2ba128480",
```
NullVoxPopuli: apparently the developertown one wasn't downloading the `dist` directory?

source/SIL.AppBuilder.Portal.Frontend/src/lib/string/unit-test.ts
```diff
@@ -0,0 +1,48 @@
+import { describe, it } from '@bigtest/mocha';
+import { expect } from 'chai';
+
+import { tokensToObject } from './utils';
+
+describe('Unit | string', () => {
```
NullVoxPopuli: @giancorzo if you need to do unit tests, you could follow this pattern.

giancorzo: +1

source/OptimaJet.DWKit.StarterApplication/OptimaJet.DWKit.StarterApplication.csproj
```diff
@@ -31,7 +31,6 @@
     <PackageReference Include="DWKit-Core" Version="2.2.3" />
     <PackageReference Include="Microsoft.ApplicationInsights.AspNetCore" Version="2.1.1" />
     <PackageReference Include="Microsoft.AspNetCore" Version="2.1.1" />
-    <PackageReference Include="Microsoft.NETCore.App" Version="2.1.1" />
```
NullVoxPopuli: this caused warnings, and apparently isn't needed

source/OptimaJet.DWKit.StarterApplication/Repositories/ProjectRepository.cs
```diff
@@ -37,6 +38,32 @@ public override IQueryable<Project> Filter(IQueryable<Project> query, FilterQuer
                 return query.FilterByOrganization(filterQuery, allowedOrganizationIds: orgIds);
             }
 
+            var value = filterQuery.Value;
+            var op = filterQuery.Operation.ToEnum<FilterOperations>(defaultValue: FilterOperations.eq);
+
+            if (filterQuery.Has(PROJECT_PRODUCT_UPDATED_DATE)) {
```
NullVoxPopuli: JADNC doesn't yet have a build in way to expose querying on relationships like this, so these custom query keys need to be interpreted in the repository

NullVoxPopuli:
This PR includes:
 - Backend support for filtering
 - Frontend upgrades
 - Frontend tests for filtering

source/SIL.AppBuilder.Portal.Frontend/src/lib/string/utils.ts
```diff
@@ -0,0 +1,51 @@
...
+    return {};
+  }
+  const tokens: string[] = search
+    .split(TOKEN_REGEX)
+    .filter(Boolean)
+    .map(t => t.replace(/(^"|"$)/g, ""));
+
+  const result: IParsedTokens = {};
+
+  for(let i = 0; i < tokens.length; i++) {
```
giancorzo: I was going to say that you could do this using a reducer (more elegant) but then I notice you need to control the index. Now that think about it... you can still skip odd indexes.

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/inputs/date-range/index.tsx
```diff
@@ -0,0 +1,91 @@
...
+import { withTranslations, i18nProps } from '@lib/i18n';
+
+import {
+  formatDate,
+  parseDate,
+} from 'react-day-picker/moment';
+
+import 'react-day-picker/lib/style.css';
+
+function tomorrow() {
```
giancorzo: I believe we have this in a lib/time.ts or we should. Also take into account timezones when doing new Date

source/SIL.AppBuilder.Portal.Frontend/src/lib/string/utils.ts
NullVoxPopuli: I can't skip odd indexes, because sometimes there isn't always a value. at the bottom of the for loop I do skip over an index IFF that next index was a value. :)

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/inputs/date-range/index.tsx
NullVoxPopuli: I see we already have _a_ implementation of tomorrow. I'll update that one with this code, and use the `@lib/date`

PR #183, opened 2018-10-08
Add buildengineproductservice

davidmoore1:


davidmoore1:
Did not check in changes to product service to automatically enqueue the request when a product is created, although I did test it that way.  Talking with Chris, we decided it would probably need for the hangfire enqueue to be done via the workflow somehow

PR #194, opened 2018-10-10
Bugfix: Form to add a new org appears in top left corner

giancorzo:
https://developertown.visualstudio.com/SIL%20International%20Scriptoria/_workitems/edit/29266

- [x] Add CSS

newcss


source/SIL.AppBuilder.Portal.Frontend/src/ui/components/header/only-logo.tsx
```diff
@@ -0,0 +1,29 @@
+import * as React from 'react';
```
NullVoxPopuli: what all is in header.scss? 
j/w because since this component is only used in the one place, maybe the file should move there?

giancorzo: To be honest, I do not understand your comment, header.css has everything need to look like the scriptoria header (few extra classes that this component won't use but nothing that will kill the bundle size).

PR #199, opened 2018-10-11
WIP: Test that setting is-public to false works correctly

chrisvire:


chrisvire:
This doesn't help at all...

PR #215, opened 2018-10-19
Save artifact modified time

davidmoore1:


source/OptimaJet.DWKit.StarterApplication/Utility/WebRequestWrapper.cs
```diff
@@ -23,6 +23,7 @@ public virtual ProductArtifact GetFileInfo(ProductArtifact artifact)
                     if (resp.ContentType != "text/html")
                     {
                         modifiedArtifact.FileSize = resp.ContentLength;
+                        modifiedArtifact.LastModified = resp.LastModified.ToUniversalTime();
```
chrisvire: Curious ... why inside `resp.ContentType != "text/html"` condition?

PR #219, opened 2018-10-19
date-range filter does not represent the selected options, because the 'last set' would clobber the first value

NullVoxPopuli:
This includes, setting the to, after setting the from,
and clearing the to disables the date-range search altogether.

The reason for this:

the query string gets translated to a single value, which makes sense -- in a key-value object, you can't have two values for one key
in the examples in jsonapi-dotnet-core, there is this:
```
?filter[attribute]=eq:value
?filter[attribute]=ne:value
?filter[attribute]=lt:value
?filter[attribute]=gt:value
?filter[attribute]=le:value
?filter[attribute]=ge:value
?filter[attribute]=like:value
?filter[attribute]=in:value
```
it would be invalid, to have all of those at once --
because key-value-ness would just take the last: `filter[attribute]=in:value`

I think this is why ransack (from rails), did stuff like
```
q[attribute_eq]=
q[attribute_ne]=
q[attribute_gt]=
```

Currently blocked until I figure out a way to filter using multiple operations on the same attribute


NullVoxPopuli:
blocked by https://github.com/orbitjs/orbit/pull/544

NullVoxPopuli:
found a bug where old filters don't get replaced :) yay

source/SIL.AppBuilder.Portal.Frontend/src/data/containers/api/with-filtering.tsx
```diff
@@ -32,6 +32,23 @@ const defaultOptions = {
...
+function withoutFilter(filters: Filter[], filter: Filter) {
+  const result = filters.filter(currentFilter => {
+    const keys = Object.keys(filter);
+    // ignore the value key, and make sure we only pull out
+    // the existing filter(s) that are the same as the target filter
+    const doesNotMatch = !keys.every(key => key === 'value' || currentFilter[key] !== filter[key]);
+
+    return doesNotMatch;
+  });
+  console.log(filters, filter, Object.keys(filter), filters[0] && Object.keys(filter).every(key => key === 'value' || filters[0][key] !== filter[key]));
```
giancorzo: remove console log

source/SIL.AppBuilder.Portal.Frontend/src/data/containers/api/with-filtering.tsx
```diff
@@ -68,29 +85,27 @@ export function withFiltering<TPassedProps>(
...
-      //       this should find and replace via attribute.
       updateFilter = (filter: IFilter) => {
         const { filters } = this.state;
 
-        const newFilters = filters.filter(currentFilter => {
-          return currentFilter.attribute !== filter.attribute;
-        });
+        const newFilters = withoutFilter(filters, filter);
+
+        console.log('newFilters, before:', newFilters);
```
giancorzo: same here, console.log

PR #225, opened 2018-10-23
update product /filter

NullVoxPopuli:
https://developertown.visualstudio.com/SIL%20International%20Scriptoria/_workitems/edit/29591

### DIFF
--- a/source/SIL.AppBuilder.Portal.Frontend/src/translations/locales/en-us.json
+++ b/source/SIL.AppBuilder.Portal.Frontend/src/translations/locales/en-us.json
@@ -156,16 +156,7 @@
     }
   },
   "productDefinitions": {
-    "filterAllProjects": "All Projects",
-    "name": {
-      "android_rab_s3": "Android RAB S3",
-      "android_dab_s3": "Android DAB S3",
-      "android_google_play": "Android Google Play",
-      "android_amazon_app": "Android Amazon App",
-      "android_scripture_earth": "Android Scripture Earth",
-      "android_s3": "Android S3",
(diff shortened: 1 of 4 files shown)


source/SIL.AppBuilder.Portal.Frontend/src/translations/locales/en-us.json
```diff
@@ -159,10 +156,15 @@
     }
   },
   "productDefinitions": {
+    "filterAllProjects": "All Projects",
     "name": {
       "android_rab_s3": "Android RAB S3",
       "android_dab_s3": "Android DAB S3",
-      "android_google_play": "Android Google Play"
+      "android_google_play": "Android Google Play",
```
giancorzo: I don't think it's a good thing to store readable names into the code, this should be handle by the backend.

NullVoxPopuli: I disagree -- because how would the backend know what language we're using? they'd be making the same API calls we'll eventually be making -- to crowdin

unless, I'm missing something?

giancorzo: Names don't need translation they are stuff name. "Android Google play" is the same in Spanish, French or other languages.

NullVoxPopuli:
@chrisvire would the product names have slightly different names in different languages?

chrisvire:
@NullVoxPopuli: product names should be static.  They shouldn't be translated.

NullVoxPopuli:
@giancorzo / @chrisvire updated

PR #243, opened 2018-10-26
eslint conversion from tslint

NullVoxPopuli:


NullVoxPopuli:
Whenever we're ready to apply these rules, we'll have a fair amount of changes:
```
✖ 6941 problems (6919 errors, 22 warnings)
  6903 errors and 0 warnings potentially fixable with the `--fix` option.
```
shouldn't take too long to apply, but we'll need to make sure all outstanding frontend work is merged

PR #245, opened 2018-10-26
I don't even know why these were passing before

NullVoxPopuli:


source/SIL.AppBuilder.Portal.Frontend/src/ui/components/user-table/row/index.tsx
```diff
@@ -67,7 +67,7 @@ class Row extends React.Component<IProps> {
             {firstName} {lastName}
           </Link>
         </td>
-        <td><RoleSelect roles={roles} user={userData} /></td>
```
NullVoxPopuli: I think this may have been causing issues

source/SIL.AppBuilder.Portal.Frontend/src/lib/routing.tsx
```diff
@@ -1,15 +1,6 @@
 import * as React from 'react';
 import { RouteComponentProps } from 'react-router';
 import { withRouter } from 'react-router-dom';
-import { withLayout } from '@ui/components/layout';
-
-export function withAuthLayout(Component) {
```
NullVoxPopuli: this was easy, and has been bugging me forever

PR #251, opened 2018-10-29
User Role management UI

NullVoxPopuli:
image

### DIFF
--- a/source/SIL.AppBuilder.Portal.Frontend/src/lib/collection.ts
+++ b/source/SIL.AppBuilder.Portal.Frontend/src/lib/collection.ts
@@ -16,3 +16,14 @@ export const applyNumberOfTimes = (x, f) => {
 
   return result;
 };
+
+export function compareVia(accessor: any) {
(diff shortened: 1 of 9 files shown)


source/SIL.AppBuilder.Portal.Frontend/src/lib/collection.ts
```diff
@@ -16,3 +16,14 @@ export const applyNumberOfTimes = (x, f) => {
 
   return result;
 };
+
+export function compareVia(accessor: any) {
```
giancorzo: what is compareVia? is that a good name?

NullVoxPopuli: did you see the usage? 
it's a comparator function :)

giancorzo: Yeah I do know what it does, just saying that name sounds vague, perhaps comparator?

PR #253, opened 2018-10-30
User groups dropdown

giancorzo:
https://developertown.visualstudio.com/SIL%20International%20Scriptoria/_workitems/edit/26029

dropdown


source/SIL.AppBuilder.Portal.Frontend/src/ui/components/user-table/row/index.tsx
```diff
@@ -54,13 +55,13 @@ class Row extends React.Component<IProps> {
...
-    const user = attributesFor(userData as any) as UserAttributes;
-    const userId = idFromRecordIdentity(userData as any);
+    const { user, t, roles, organizations } = this.props;
+    const { givenName, familyName, isLocked } = attributesFor(user);
+    const userId = idFromRecordIdentity(user);
 
-    const firstName = user.givenName || `(${t('profile.firstName')})`;
-    const lastName = user.familyName || `(${t('profile.lastName')})`;
-    const isActive = !user.isLocked;
+    const firstName = givenName || `(${t('profile.firstName')})`;
```
NullVoxPopuli: we're actually supposed to use `name` now, because first/last isn't always in that order

PR #254, opened 2018-10-31
scope the row to the current user's organizations

NullVoxPopuli:


source/SIL.AppBuilder.Portal.Frontend/src/ui/components/user-table/row/active-roles-display.tsx
```diff
@@ -0,0 +1,66 @@
...
+  render() {
+    const { t } = this.props;
+    const activeRoles = this.activeRoles();
+
+    if (activeRoles.length === 0) {
+      return t('users.noRoles');
+    }
+
+    return activeRoles.map(( role, i ) => {
+      return <span key={i}>{attributesFor(role).roleName}</span>;
```
giancorzo: no need for the inner return if you remove the { }

NullVoxPopuli: that makes debugging more challenging. :)

PR #260, opened 2018-11-01
Don't remove group membership if project exists for that user and group

davidmoore1:


source/OptimaJet.DWKit.StarterApplication/Forms/GroupMemberships/DeleteForm.cs
```diff
@@ -0,0 +1,52 @@
...
+            {
+                var message = "Record being deleted not found";
+                AddError(message, 404);
+            }
+            else
+            {
+                var projects = ProjectRepository.Get()
+                                                .Where(p => p.GroupId == groupMembership.GroupId && p.OwnerId == groupMembership.UserId)
+                                                .AsEnumerable();
+                if (projects.Count() > 0)
```
chrisvire: You could use `.Any()` instead of getting the enumerable list and then getting the count.

PR #265, opened 2018-11-03
New Project Group UX Fixes + Tests

NullVoxPopuli:


source/SIL.AppBuilder.Portal.Frontend/src/ui/components/user-table/row/multi-role-select/display.tsx
```diff
@@ -0,0 +1,83 @@
...
+      <Dropdown
+        data-test-role-multi-select
+        multiple
+        text={roleNames}
+        className='w-100 multiDropdown'
+      >
+        <Dropdown.Menu data-test-role-menu>
+          {
+            organizations.map((organization, index) => {
+              const organizationName = attributesFor(organization).name;
```
giancorzo: const organizationName = attributesFor(organization).name **|| ''**; //just in case?

NullVoxPopuli: naw, if we need to do that, we aren't handling asynchrony properly

PR #277, opened 2018-11-12
Fix label color

giancorzo:
p2
p1


NullVoxPopuli:
the outline needs to be blue for the checkboxes.

giancorzo:
new

PR #278, opened 2018-11-12
Bugfix: "Archive" option in kabob updates

giancorzo:
https://developertown.visualstudio.com/SIL%20International%20Scriptoria/_workitems/edit/29891

directory
myown


source/SIL.AppBuilder.Portal.Frontend/src/ui/components/project-table/table/row/row-actions.tsx
```diff
@@ -2,60 +2,72 @@ import * as React from 'react';
...
     const { dateArchived } = attributesFor(project);
 
     const dropdownItemText = !dateArchived ?
       t('project.dropdown.archive') :
       t('project.dropdown.reactivate');
 
+    const requireRoleProps = {
+      roleName: ROLE.OrganizationAdmin,
+      overrideIf: (props: {}) => {
+        return owner.id === currentUser.id;
```
NullVoxPopuli: owner and currentUser should come from the props of overrideIf -- as is, there is some goofy scope cross-over/leakage here.

PR #279, opened 2018-11-13
Bugfix: not hidding server errors 

giancorzo:
I remove the custom error message and show what the server is returning.

### DIFF
--- a/source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/projects/show/with-project-operations.tsx
+++ b/source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/projects/show/with-project-operations.tsx
@@ -37,11 +37,7 @@ export function withProjectOperations(WrappedComponent) {
         );
       } catch (e) {
         console.error(e);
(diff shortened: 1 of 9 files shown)


source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/projects/show/with-project-operations.tsx
```diff
@@ -37,11 +37,7 @@ export function withProjectOperations(WrappedComponent) {
         );
       } catch (e) {
         console.error(e);
-        toast.error(
-          !dateArchived ?
-            t('project.operations.archive.error') :
-            t('project.operations.reactivate.error')
-        );
+        toast.error(e);
```
NullVoxPopuli: does toast.error know how to handle `Error` objects?

giancorzo: it does. Most likely because it calls toString of Error object

PR #281, opened 2018-11-13
Superadmins can create/update records outside of their organization

davidmoore1:


source/OptimaJet.DWKit.StarterApplication/Forms/BaseForm.cs
```diff
@@ -58,6 +64,15 @@ protected void ValidateOrganizationHeader(int initialOrganizationId, string obje
                 }
             }
         }
+        protected bool IsCurrentUserSuperAdmin()
```
NullVoxPopuli: would it be worth it to put this closer to where we get the current user?
or maybe have another service that abstracts this away?
we may need it in general, or helpers like it for more fine-grain permission controls.

I'm totes fine with that getting thrown in the backlog though, to be abstracted when needed.

davidmoore1: Added backlog item 109 to refactor this once it became clearer who all would need it and whether other generalization needed to be added.

PR #288, opened 2018-11-16
Add Mechanism to initialize the BuildEngineApiAccessToken for CI

chrisvire:
For CI and Dev environments, we bootstrap regularly and need to have Organization.BuildEngineApiAccessToken updated.  This change enqueues a job at startup if `SAMPLEDATA_BUILDENGINE_API_ACCESS_TOKEN` is set.

set `SAMPLEDATA_BUILDENGINE_API_ACCESS_TOKEN=token` to change them all
set `SAMPLEDATA_BUILDENGINE_API_ACCESS_TOKEN=1=token1|2=token2|3=token3` to change them based on the Organization.Id (1=SIL, 2=DT, 3=Kalaam)

For development, set this in source/OptimaJet.DWKit.StarterApplication/.env (won't be checked in).

This variable will be set in TerraForm for CI environment so that the values are updated when the CI environment is reset.

### DIFF
--- a/source/OptimaJet.DWKit.StarterApplication/Utility/ApplicationExtensions.cs
+++ b/source/OptimaJet.DWKit.StarterApplication/Utility/ApplicationExtensions.cs
@@ -10,6 +10,7 @@ using OptimaJet.DWKit.Core;
 using OptimaJet.DWKit.StarterApplication.Services.BuildEngine;
 using OptimaJet.DWKit.StarterApplication.Services.Workflow;
 using OptimaJet.Workflow.Core.Runtime;
+using static OptimaJet.DWKit.StarterApplication.Utility.EnvironmentHelpers;
 
 namespace OptimaJet.DWKit.StarterApplication.Utility
 {
@@ -38,6 +39,11 @@ namespace OptimaJet.DWKit.StarterApplication.Utility
 
         public static IApplicationBuilder UseBuildEngine(this IApplicationBuilder app, IConfigurationRoot configuration)
         {
+            var sampleDataApiToken = GetVarOrDefault("SAMPLEDATA_BUILDENGINE_API_ACCESS_TOKEN", String.Empty);
+            if (!String.IsNullOrEmpty(sampleDataApiToken))
+            {
+                BackgroundJob.Enqueue<BuildEngineSystemMonitor>(service => service.SetSampleDataApiToken(sampleDataApiToken));
+            }
             RecurringJob.AddOrUpdate<BuildEngineSystemMonitor>("BuildEngineMonitor", service => service.CheckBuildEngineStatus(), Cron.MinuteInterval(5));
 
             return app;
--- a/source/OptimaJet.DWKit.StarterApplication/Services/BuildEngine/BuildEngineSystemMonitor.cs
+++ b/source/OptimaJet.DWKit.StarterApplication/Services/BuildEngine/BuildEngineSystemMonitor.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Collections.Generic;
 using System.Linq;
 using JsonApiDotNetCore.Data;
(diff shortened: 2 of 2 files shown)


NullVoxPopuli:
For the production version of this are we going to have a default build engine?

chrisvire:
@NullVoxPopuli, for production, we will be using https://buildengine.lsdev.sil.org.  However, we won't be using the sample data. So this is irrelevant to production.

Actually, setting the Env Var in Travis is useless. I need to set it in TerraForm for the staging environment.

source/OptimaJet.DWKit.StarterApplication/Utility/ApplicationExtensions.cs
```diff
@@ -38,6 +39,11 @@ public static IApplicationBuilder UseWorkflow(this IApplicationBuilder app, ICon
 
         public static IApplicationBuilder UseBuildEngine(this IApplicationBuilder app, IConfigurationRoot configuration)
         {
+            var sampleDataApiToken = GetVarOrDefault("SAMPLEDATA_BUILDENGINE_API_ACCESS_TOKEN", String.Empty);
```
NullVoxPopuli: I just mean that this behavior doesn't seem tied to sample data, but more just a default / override.

like, for production, we could have the exact same code, provided we use the phrasing:

```cs
var defaultToken = GetVarOrDefault("DEFAULT_BUILDENGINE_API_ACCESS_TOKEN", String.Empty);

if (!String.IsNullOrEmpty(defaultToken))
{
    // somewhere in the BuildEngine Code, a method should be abstracted for accessing the API token
    // which would either get the set organization token, or the default token from here.
    // and, actually, if the BuildEngine stuff is somewhat stateless w/r/t the api token acces, 
    // these code changes could be remove entirely, and we could just do the abstracted method for api token 
    // access to either be the org specified token or the DEFAULT_BUILDENGINE_API_ACCESS_TOKEN
    BackgroundJob.Enqueue<BuildEngineSystemMonitor>(service => service.SetApiToken(defaultToken));
}
```

NullVoxPopuli: how do you imagine it work in production?

chrisvire: For BuildEngine, we have an admin interface for creating them.  I believe they were supposed to be filled in by the SuperAdmin in the SuperAdmin UI.

NullVoxPopuli: ah, so you just don't want a default in production? (this is fine, I'm just trying to figure out what the intent is)

chrisvire: "For CI and Dev environments, we bootstrap regularly and need to have Organization.BuildEngineApiAccessToken updated."

NullVoxPopuli: right, but that just felt narrow to me, leaving production out of the picture. Apologies if I wasn't clear in expressing my confusion.

If the intent is to:

 - never have a default build engine configuration in production
 - it is up to the super admin to enable the default build engine via manual configuration

that's fine, I'd just like to know if that _is_ the case

chrisvire: Yep.  This is just for Dev and CI. Not for production. For Production, we already have this exposed through the Organization UI.  I forgot we had already done this.  No need for SuperAdmin UI.

NullVoxPopuli: aight. this PR is :+1: then. thanks for explaining!

PR #289, opened 2018-11-16
Enforce roles on access of users and org settings screens

NullVoxPopuli:
https://developertown.visualstudio.com/SIL%20International%20Scriptoria/_workitems/edit/29889

App builders do not see Manage Users in side nav
App builders do not see Org Settings in side nav
Org admins can navigate to both manage users and org settings


NullVoxPopuli:
first test results are fake.

looks like in one of the times we merged frontend code, we forgot to remove `describe.only`

@giancorzo do you think you could start using the debug UI and click into which tests you are debugging instead of using only? thanks!!!

giancorzo:
Yes I can, sorry about that. I always use debug UI problem is that we have 200+ test :(

NullVoxPopuli:
@giancorzo ready for review

PR #303, opened 2018-11-27
bug-29865

jonnolen:
made re-entrant calls to item onSelectionChange with the same item have no effect until after the first update is complete.


source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/projects/show/products/index.tsx
```diff
@@ -29,21 +29,34 @@ interface IOwnProps {
...
 
-    const { t, updateProduct } = this.props;
-    try {
-      await updateProduct(item);
-      toast.success(t('updated'));
-    } catch (e) {
-      toast.error(e.message);
+  onSelectionChange = async (item: ProductDefinitionResource) => {
+    if (!this.pendingUpdates[item.id]){
+      this.pendingUpdates[item.id] = true;
```
NullVoxPopuli: nice use of of state outside of the recommended state management! :)

using setState triggers re-renders, so this is perfect.

NullVoxPopuli:
Should probably add a test though

source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/projects/show/products/index.tsx
```diff
@@ -29,21 +29,34 @@ interface IOwnProps {
...
+  [itemId: string]: boolean;
+}
+
 type IProps =
   & IOwnProps
   & i18nProps
   & IDataActionsProps;
 
 class Products extends React.Component<IProps> {
+  pendingUpdates: IPendingUpdates;
```
NullVoxPopuli: idk what conventions you prefer, or whatever, but I guess just fyi, this can be one-lined to

```ts
pendingUpdates: IPendingUpdates = {};
```

and then the constructor can go away

source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/projects/show/products/index.tsx
jonnolen: yep.

source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/projects/show/products/index.tsx
jonnolen: i'll one line it.

jonnolen:
i'll take a stab at a test.

PR #304, opened 2018-11-27
Fix exceptions caused by null build engine url or token key

davidmoore1:


source/OptimaJet.DWKit.StarterApplication/Services/BuildEngine/BuildEngineServiceBase.cs
```diff
@@ -35,9 +35,14 @@ protected bool BuildEngineLinkAvailable(Organization organization)
 
             return systemStatus.SystemAvailable;
         }
-        protected void SetBuildEngineEndpoint(Organization organization)
+        protected bool SetBuildEngineEndpoint(Organization organization)
         {
+            if (organization.BuildEngineUrl is null)
```
chrisvire: Need BuildEngineApiAccessToken too

source/OptimaJet.DWKit.StarterApplication/Services/BuildEngine/BuildEngineSystemMonitor.cs
```diff
@@ -114,6 +114,10 @@ public void SetSampleDataApiToken(string sampleDataApiToken)
 
         private bool CheckConnection(SystemStatus systemEntry)
         {
+            if (systemEntry.BuildEngineUrl is null)
+            {
```
chrisvire: Need BuildEngineApiAccessToken too

PR #305, opened 2018-11-28
Bugfix: When a project is archived it was not been filtered out until you refresh the page

giancorzo:
https://developertown.visualstudio.com/SIL%20International%20Scriptoria/_workitems/edit/29887

refresh


NullVoxPopuli:
The gif is a nice touch!

NullVoxPopuli:
who's this? 
image

source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/project-directory/list/__tests__/pagination-acceptance-test.tsx
```diff
@@ -54,7 +54,7 @@ describe('Acceptance | Pagination', () => {
 
         navigateTo(screen.path);
 
-        it('Pagination is not displayed',() => {
+        xit('Pagination is not displayed',() => {
```
NullVoxPopuli: why are these disabled?

source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/projects/list/my-projects.tsx
```diff
@@ -36,9 +36,10 @@ export default compose(
     };
   }),
   withNetwork(),
+  withCache( () =>  ( { projects :  q => q.findRecords( PROJECT ) }) ),
```
NullVoxPopuli: some strange formatting going on in here

source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/projects/list/my-projects.tsx
```diff
@@ -36,9 +36,10 @@ export default compose(
...
   withLoader(({ error, projects }) => !error && !projects),
   withProps(({ projects }) => ({
-    projects: projects.filter(resource => resource.type === PROJECT),
+    projects: projects.filter(resource => resource.type === PROJECT && resource.attributes.dateArchived == null),
```
NullVoxPopuli: ez :)

giancorzo:
> who's this?
> image

Luis is a Tekton dev that is helping me out with the project while I'm traveling.

source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/project-directory/list/__tests__/pagination-acceptance-test.tsx
giancorzo: Because pagination is never hidden, you said some time ago that because of a bug in the JSON API we couldn't hide the pagination. So I don't how this test even worked.

NullVoxPopuli: well, we can't just make changes without understanding why. :)

It's either was working due to a timing issue, or because the scenario / setup, is, indeed correct.

There are 0 projects -> Pagination is not displayed -- this makes sense.
There is less than one page worth of projects -> Pagination is not displayed -- also makes sense

source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/projects/list/my-projects.tsx
giancorzo: what is ez?

NullVoxPopuli: this way of filtering out archived things :)

source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/project-directory/list/__tests__/pagination-acceptance-test.tsx
NullVoxPopuli: no, if it passes right now, just leave it enabled?

giancorzo: yeah, but they are not passing, that is why I change them.

giancorzo: Something is wrong with this test because as you can see on the screen it is passing when the pagination is been there

bug

giancorzo: Ok I found why is working in Master, that test is using isPresent to check if the pagination is there but what happen is that the table is not fully loaded when you enter to that test so is present return false correctly. If you use isVisible instead of isPresent the test breaks.

NullVoxPopuli: So it sounds like the 'it' should be 'it.always'
And if the behavior isn't implemented, we can only allow it to be commented out with the proper documentation comments surrounding those tests. To protect future people from investigating again.

PR #308, opened 2018-11-30
bugfix: Source of error message displayed when trying to create a new project

giancorzo:
https://developertown.visualstudio.com/SIL%20International%20Scriptoria/_workitems/edit/30270

bugmessage

### DIFF
--- a/source/SIL.AppBuilder.Portal.Frontend/src/ui/components/errors/parse-error.ts
+++ b/source/SIL.AppBuilder.Portal.Frontend/src/ui/components/errors/parse-error.ts
@@ -10,7 +10,7 @@ export function parseError(error: any): ParsedError {
   if( errorDetail ) {
     return {
       title: errorDetail
-    }
+    };
   }
   if (error instanceof RecordNotFoundException) {
     return {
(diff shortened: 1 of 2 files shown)


source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/projects/new/display.tsx
```diff
@@ -75,7 +75,8 @@ export default class Display extends React.Component<IProps, IState> {
       history.push(`/projects/${id}`);
 
     } catch (e) {
-      toast.error(t('errors.generic', { errorMessage: e.message }));
+      const message = (e.data && e.data.errors && e.data.errors.lenght > 0 && e.data.errors[0].detail) || e.message;
```
NullVoxPopuli: this logic should go into parse-error.ts

also, you have a typo on length

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/errors/parse-error.ts
```diff
@@ -6,6 +6,12 @@ export interface ParsedError {
 }
 
 export function parseError(error: any): ParsedError {
+  const errorDetail = error.data && error.data.errors && error.data.errors.length > 0 && error.data.errors[0].detail;
```
NullVoxPopuli: This should be put after the instanceof checks.
otherwise you're totally ignoring the more specific error-handling.

PR #309, opened 2018-11-30
Bugfix: Change text on toast

giancorzo:
https://developertown.visualstudio.com/SIL%20International%20Scriptoria/_workitems/edit/30256

screen shot 2018-11-30 at 11 43 28 am


source/SIL.AppBuilder.Portal.Frontend/src/translations/locales/en-us.json
```diff
@@ -218,7 +218,7 @@
       "unpublished": "Unpublished",
       "noArtifacts": "No artifacts",
       "rebuild": "rebuild",
-      "creationInProgress": "There are server resources being created for this project.  Products cannot be added yet. Please refresh in a few minutes."
+      "creationInProgress": "You cannot add a product until the project setup has been completed. Please refresh in a while."
```
NullVoxPopuli: Do we need to have "Please refresh in a while"?

this is a really long message.

maybe it'd be better to prefix with `Pending: `

PR #310, opened 2018-11-30
Build History

chrisvire:
* Add `ProductBuild` table for history of builds per product
* Set Version of ProductBuild based on artifact with `ArtifactType=version` and `ContentType=application/json`
* `Product.WorkflowBuildId` follows the most recent `ProductBuild`

### DIFF
--- a/source/OptimaJet.DWKit.StarterApplication/Services/BuildEngine/BuildEngineBuildService.cs
+++ b/source/OptimaJet.DWKit.StarterApplication/Services/BuildEngine/BuildEngineBuildService.cs
@@ -236,7 +236,7 @@ namespace OptimaJet.DWKit.StarterApplication.Services.BuildEngine
             await ProductArtifactRepository.CreateAsync(updatedArtifact);
 
 #pragma warning disable RECS0061 // Warns when a culture-aware 'EndsWith' call is used by default.
-            if (!String.IsNullOrEmpty(value) && value.EndsWith("version.json"))
+            if (key == "version" && updatedArtifact.ContentType == "application/json")
 #pragma warning restore RECS0061 // Warns when a culture-aware 'EndsWith' call is used by default.
             {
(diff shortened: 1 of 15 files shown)


source/OptimaJet.DWKit.StarterApplication/Services/BuildEngine/BuildEngineBuildService.cs
```diff
@@ -9,29 +9,37 @@
...
             WebRequestWrapper webRequestWrapper,
+            IWebClient webClient,
             IJobRepository<Product, Guid> productRepository,
             IJobRepository<ProductArtifact> productArtifactRepository,
+            IJobRepository<ProductBuild> productBuildRepository,
             IJobRepository<SystemStatus> systemStatusRepository
         ) : base(buildEngineApi, systemStatusRepository)
         {
             RecurringJobManager = recurringJobManager;
             WebRequestWrapper = webRequestWrapper;
```
chrisvire: Eventually, I would like to see WebRequestWrapper be more generic (like IWebClient) instead of being specific to ProductArtifact.

source/OptimaJet.DWKit.StarterApplication/Utility/SystemWebClient.cs
```diff
@@ -0,0 +1,10 @@
+﻿using System;
+using System.Net;
+
+namespace OptimaJet.DWKit.StarterApplication.Utility
+{
+    public class SystemWebClient : WebClient, IWebClient
+    {
+
```
davidmoore1: So this is here just so you can scope it?

chrisvire: @davidmoore1: Sorry, I didn't see the comment.  This is here so that it can be brought in as an injected dependency that can be mocked in the tests.

PR #312, opened 2018-12-03
Bugfix: Assign user to a group

giancorzo:
user-group


source/SIL.AppBuilder.Portal.Frontend/src/data/containers/with-relationship.tsx
```diff
@@ -100,7 +105,6 @@ export function withRelationships<T>(mappingFn: (props: T) => MapFnResult) {
           relationshipsToFind: mapResult,
         };
       }),
-      withOrbit({}),
```
NullVoxPopuli: why was this removed?

PR #315, opened 2018-12-05
Various Bugfixes

NullVoxPopuli:
Change text: 
https://developertown.visualstudio.com/SIL%20International%20Scriptoria/_workitems/edit/30369

An actual bug: 
https://developertown.visualstudio.com/SIL%20International%20Scriptoria/_workitems/edit/30401
(this is what the tests are for)

Bonus Bug:
the Logos were super bothering me.............
So they contain themselves now.
I then noticed the sidebar image was trash:
image

So, that's been fixed as well

image is smaller than container (not blown up)
image
image: https://www.developertown.com/wp-content/uploads/2018/01/DT-PoundHouse-Blue.png

image is larger than container (needs to shrink)
image
image
and got rid of the background: 
image
image
image: https://pbs.twimg.com/profile_images/634058281472782336/_4-1Li0w_400x400.png

image is much wider than container (needs to shrink)
image
image
image

image: https://www.developertown.com/wp-content/uploads/2018/01/DT-Logo-Horizontal-Normal-LARGE.png

and with no image: 
image
image
image

### DIFF
--- a/source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/organizations/settings/index.tsx
+++ b/source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/organizations/settings/index.tsx
@@ -157,7 +157,9 @@ export default compose(
   })),
   withRole(ROLE.OrganizationAdmin, {
     redirectTo: '/',
-    checkOrganizationOf: (props: IProps) => props.organization
+    forOrganization: (props: IProps) => {
+      return props.organization;
+    }
   }),
   withDataActions,
   withTranslations,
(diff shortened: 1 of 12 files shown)


source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/organizations/settings/index.tsx
```diff
@@ -157,7 +157,9 @@ export default compose(
   })),
   withRole(ROLE.OrganizationAdmin, {
     redirectTo: '/',
-    checkOrganizationOf: (props: IProps) => props.organization
+    forOrganization: (props: IProps) => {
```
NullVoxPopuli: this is all that was needed to fix bug https://developertown.visualstudio.com/SIL%20International%20Scriptoria/_workitems/edit/30401

giancorzo: interesting...

NullVoxPopuli: yeah, an organization doesn't belong to an organization, so that's where the id of undefined error was coming from

PR #319, opened 2018-12-05
Bugfix/various timing request errors

NullVoxPopuli:
Related:https://developertown.visualstudio.com/SIL%20International%20Scriptoria/_workitems/edit/30359

Probably others

### DIFF
--- a/source/SIL.AppBuilder.Portal.Frontend/package.json
+++ b/source/SIL.AppBuilder.Portal.Frontend/package.json
@@ -63,7 +63,7 @@
     "react-dropzone-component": "^3.2.0",
     "react-i18next": "^8.3.8",
     "react-notify-toast": "^0.5.0",
-    "react-orbitjs": "github:NullVoxPopuli/react-orbitjs#869efe1bec6fde7593e6a64a523c0e4c4f1df0da",
+    "react-orbitjs": "github:NullVoxPopuli/react-orbitjs#c1f47d604a56be00a2986740ca841181db225f09",
     "react-redux": "^5.0.7",
     "react-router": "^4.3.1",
     "react-router-dom": "^4.3.1",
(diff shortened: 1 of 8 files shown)


source/SIL.AppBuilder.Portal.Frontend/src/data/query.tsx
```diff
@@ -93,12 +93,8 @@ export function queryApi<T>(mapRecordsToProps, options?: IQueryOptions) {
           return queryResult;
         });
 
-        try {
-          await Promise.all(requestPromises);
```
NullVoxPopuli: the main issue is that after a 500 error, all promises seem to never resolve

NullVoxPopuli: this was because the requestQueue is halted on a 500 to give us a chance to handle the error if we need to

source/SIL.AppBuilder.Portal.Frontend/src/lib/promises.ts
```diff
@@ -0,0 +1,13 @@
+export function timeoutablePromise(timeoutMs, promise) {
+  const timeout = new Promise((resolve, reject) => {
+    const id = setTimeout(() => {
+      clearTimeout(id);
+      reject(`Timed out after ${timeoutMs} ms.`);
```
NullVoxPopuli: I don't think this needs translating. it *should* never happen and is really for our (the devs) benefit

source/SIL.AppBuilder.Portal.Frontend/package.json
```diff
@@ -63,7 +63,7 @@
     "react-dropzone-component": "^3.2.0",
     "react-i18next": "^8.3.8",
     "react-notify-toast": "^0.5.0",
-    "react-orbitjs": "github:NullVoxPopuli/react-orbitjs#869efe1bec6fde7593e6a64a523c0e4c4f1df0da",
+    "react-orbitjs": "github:NullVoxPopuli/react-orbitjs#c1f47d604a56be00a2986740ca841181db225f09",
```
NullVoxPopuli: some performance improvements

source/SIL.AppBuilder.Portal.Frontend/src/lib/promises.ts
giancorzo: agree, just a suggestion invert the properties and give it a default 

timeoutablePromise(promise, timeoutMs=5000)

NullVoxPopuli: if we use this again, we can give it a default :)

PR #321, opened 2018-12-06
Bugfix/update product data model

NullVoxPopuli:
image

Latest Build: 
image

image

### DIFF
--- a/source/SIL.AppBuilder.Portal.Frontend/package.json
+++ b/source/SIL.AppBuilder.Portal.Frontend/package.json
@@ -173,5 +173,9 @@
     "webpack-dev-server": "^3.1.4",
     "webpack-node-externals": "^1.7.2",
     "whatwg-fetch": "^3.0.0"
+  },
+  "resolutions": {
+    "react": "^16.0.0",
+    "react-dom": "^16.0.0"
   }
 }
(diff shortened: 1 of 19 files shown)


source/SIL.AppBuilder.Portal.Frontend/package.json
```diff
@@ -173,5 +173,9 @@
     "webpack-dev-server": "^3.1.4",
     "webpack-node-externals": "^1.7.2",
     "whatwg-fetch": "^3.0.0"
+  },
+  "resolutions": {
```
NullVoxPopuli: This ensures we only have one copy of react + react-dom in our build.

Bundle Analysis:

image

auth0-lock is currently: 103KB gzipped, and before it was 173KB gzipped

source/SIL.AppBuilder.Portal.Frontend/src/ui/styles/semantic-ui-overrides.scss
```diff
@@ -180,6 +180,14 @@ Dropdown Menu
   }
 }
 
+.ui.dropdown.is-large {
```
giancorzo: from the name of the class I expected width:100% or something like that... but it's more related to font size large

PR #324, opened 2018-12-07
Send organization invite emails to all super admins

davidmoore1:
Don't use environment variable

### DIFF
--- a/source/OptimaJet.DWKit.StarterApplication/Models/User.cs
+++ b/source/OptimaJet.DWKit.StarterApplication/Models/User.cs
@@ -87,12 +87,9 @@ namespace OptimaJet.DWKit.StarterApplication.Models
             {
                 locale = Locale;
             }
-            else
+            else if ((CultureInfo.CurrentCulture != null) && !String.IsNullOrEmpty(CultureInfo.CurrentCulture.Name))
             {
-                if ((CultureInfo.CurrentCulture != null) && !String.IsNullOrEmpty(CultureInfo.CurrentCulture.Name))
-                {
-                    locale = CultureInfo.CurrentCulture.Name;
-                }
+                locale = CultureInfo.CurrentCulture.Name;
             }
             return locale;
         }
--- a/source/OptimaJet.DWKit.StarterApplication/Models/User.cs
+++ b/source/OptimaJet.DWKit.StarterApplication/Models/User.cs
@@ -80,10 +80,14 @@ namespace OptimaJet.DWKit.StarterApplication.Models
             return userRole != null;
         }
 
-        [NotMapped]
-        public string LocaleOrDefault
+        public string LocaleOrDefault()
         {
-            get => Locale ?? (CultureInfo.CurrentCulture.Name ?? "en-US");
+            var locale = Locale ?? ((CultureInfo.CurrentCulture == null) ? "en-US" : CultureInfo.CurrentCulture.Name ?? "en-US");
+            if (String.IsNullOrEmpty(locale))
+            {
+                locale = "en-US";
+            }
+            return locale;
         }
     }
 }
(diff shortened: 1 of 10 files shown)


source/OptimaJet.DWKit.StarterApplication/OptimaJet.DWKit.StarterApplication.csproj
```diff
@@ -20,6 +20,10 @@
 
   <PropertyGroup Condition=" '$(RunConfiguration)' == 'Default' ">
     <StartAction>Project</StartAction>
+    <EnvironmentVariables>
+      <Variable name="ASPNETCORE_ENVIRONMENT" value="Development" />
+      <Variable name="ASPNETCORE_ENVIRONMENT" value="Development" />
+    </EnvironmentVariables>
```
chrisvire: Please don't add these.

source/OptimaJet.DWKit.StarterApplication/Models/User.cs
```diff
@@ -80,10 +80,14 @@ public class User : Identifiable, ITrackDate
...
-        public string LocaleOrDefault
+        public string LocaleOrDefault()
         {
-            get => Locale ?? (CultureInfo.CurrentCulture.Name ?? "en-US");
+            var locale = Locale ?? ((CultureInfo.CurrentCulture == null) ? "en-US" : CultureInfo.CurrentCulture.Name ?? "en-US");
+            if (String.IsNullOrEmpty(locale))
+            {
+                locale = "en-US";
+            }
+            return locale;
```
chrisvire: There are a lot of "en-US" in this.  Is there a way to simplify?

source/OptimaJet.DWKit.StarterApplication/Models/User.cs
```diff
@@ -80,10 +80,21 @@ public class User : Identifiable, ITrackDate
...
         {
-            get => Locale ?? (CultureInfo.CurrentCulture.Name ?? "en-US");
+            var locale = "en-US";
+            if (!String.IsNullOrEmpty(Locale))
+            {
+                locale = Locale;
+            }
+            else
+            {
+                if ((CultureInfo.CurrentCulture != null) && !String.IsNullOrEmpty(CultureInfo.CurrentCulture.Name))
```
chrisvire: do
```
else if (...)
{
}
```
instead of 
```
else 
{
   if (...)
   {
   }
}  
```

source/OptimaJet.DWKit.StarterApplication/Services/OrganizationInviteRequestService.cs
```diff
@@ -10,42 +13,55 @@ namespace OptimaJet.DWKit.StarterApplication.Services
...
         public void Process(OrganizationInviteRequestServiceData data)
         {
             var request = requestRepository.GetAsync(data.Id).Result;
-            var email = new Email
+            var superAdmins = userRolesRepository.Get()
+                .Include(ur => ur.User)
+                .Include(ur => ur.Role)
+                .Where(ur =>  ur.Role.RoleName == RoleName.SuperAdmin
+                       && ur.User.EmailNotification == true
+                )
```
chrisvire: I am wondering if the `EmailNotification` should only apply to AppBuilder role and not SuperAdmin (and OrgAdmin?) role. To me, it seems like we don't want SuperAdmin to opt out of email since these are things they need to deal with.

source/OptimaJet.DWKit.StarterApplication/Models/User.cs
davidmoore1: Done

source/OptimaJet.DWKit.StarterApplication/Services/OrganizationInviteRequestService.cs
chrisvire: After we talked about it, we shouldn't have SuperAdmin or OrgAdmin opt out of email for now.

PR #329, opened 2018-12-11
Update to DotNet SDK 2.1.500 and Runtime 2.1.6

chrisvire:
* Update DotNet from 2.1.402 to 2.1.500
* Change runtime-release from sdk-bionic to aspnetcore-runtime-alpine


chrisvire:
From 2.4GB to 300MB, if I remember correctly

PR #330, opened 2018-12-11
Modify Notifications Table

davidmoore1:
This pull request modifies the notification table to contain the translated message instead of the message id and substitutions.  It also changes the table to include a flag as to whether an email needs to be sent.  Changes were made to the send notification service to check that flag and to accept a dictionary with the parameters for the message to accommodate those changes

### DIFF
--- a/source/OptimaJet.DWKit.StarterApplication/Models/Notification.cs
+++ b/source/OptimaJet.DWKit.StarterApplication/Models/Notification.cs
@@ -9,9 +9,6 @@ namespace OptimaJet.DWKit.StarterApplication.Models
 {
     public class Notification : Identifiable, ITrackDate
     {
-        [Attr("message-id")]
-        public String MessageId { get; set; }
-
         [HasOne("user", Link.None)]
         public virtual User User { get; set; }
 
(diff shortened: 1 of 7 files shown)


source/OptimaJet.DWKit.StarterApplication/Models/Notification.cs
```diff
@@ -29,19 +26,10 @@ public class Notification : Identifiable, ITrackDate
...
-                : JsonConvert.SerializeObject(MessageSubstitutions);
-            set => MessageSubstitutions = string.IsNullOrWhiteSpace(value)
-                ? new Dictionary<string, object>()
-                : JObject.Parse(value).ToObject<Dictionary<string, object>>();
-        }
+        [Attr("message")]
+        public String Message { get; set; }
 
+        [Attr("send-email")]
+        public bool SendEmail { get; set; } = true;
```
chrisvire: If you want the default to be true, then you also need a default in `AppDbContent.OnModelCreating`.  Otherwise, other mechanisms that could create a Notification (JsonApi or non-EF database interface) will not have the correct default.

PR #333, opened 2018-12-11
WIP: Fix issue where selected organization does not persist between refreshes

NullVoxPopuli:


source/SIL.AppBuilder.Portal.Frontend/src/data/containers/with-current-organization/hoc.ts
```diff
@@ -0,0 +1,61 @@
+import { withData, WithDataProps } from 'react-orbitjs';
+import { connect } from 'react-redux';
+import {
+  compose, withProps, branch, renderComponent
+} from 'recompose';
+
+import { State } from '@store/reducers';
```
giancorzo: something else to add to linting... we should change that to IState to not get confuse

NullVoxPopuli: aye, I'll make a PBI

NullVoxPopuli: https://developertown.visualstudio.com/SIL%20International%20Scriptoria/_workitems/edit/30645

source/SIL.AppBuilder.Portal.Frontend/src/lib/collection.ts
```diff
@@ -40,3 +40,10 @@ export function compareVia(accessor: any, invert?: boolean) {
     return 0;
   };
 }
+
+// This is a stupid way to 'deeply' compare things.
+// But it kinda works.
+// Functions are omitted from the comparison
+export function arePropsEqual(a, b) {
```
giancorzo: perhaps a better name... because we could eventually use it to compare objects that are not props

NullVoxPopuli: yeah, I just copy pasted this from the existing location

PR #334, opened 2018-12-12
Super admin adds product description

giancorzo:
https://developertown.visualstudio.com/SIL%20International%20Scriptoria/_workitems/edit/29618

Note: we don't have super admin yet to create product definitions so product creation needs to happen in the database.

modal-product
producst
stores


NullVoxPopuli:
multi-select is used in a few places, can you post screenshots of all the other places its used?

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/product-icon/index.tsx
```diff
@@ -11,31 +11,44 @@ const colorStyles = {
...
+  };
+};
+
 const DEFAULT_COLOR = { color: '#9b9b9b' };
 
 const iconMap = {
-  android: (selected) => <AndriodIcon style={selected ? colorStyles.android : DEFAULT_COLOR} />,
-  html: (selected) => <WebIcon style={selected ? colorStyles.html : DEFAULT_COLOR} />,
+  android: (selected, size) =>
+    <AndriodIcon style={styles('android',selected, size)} />,
```
davidmoore1: Not that this is a big thing, but shouldn't this be AndroidIcon not AndriodIcon

giancorzo: nice catch

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/sidebar/navigation.tsx
```diff
@@ -91,7 +91,7 @@ class Navigation extends React.Component<IProps & i18nProps> {
           <RequireRole roleName={ROLE.OrganizationAdmin}>
             <MenuItem
               name={t('sidebar.organizationSettings')}
-              to={`/organizations/${currentOrganizationId}/settings`}
+              to={`/organizations/${currentOrganizationId}/settings/basic-info`}
```
giancorzo: this is not related to this issue, but a new bug that I found while navigation to organization settings. This happened because I changed the Sidebar menu for super admin.

NullVoxPopuli: what was the bug?

NullVoxPopuli: so what happens when you visit `organizations/id/settings/`?

giancorzo: several routes get mark as active because in responsiveNav I changed it to
`isActive={() => location.pathname.startsWith(i.to)}` because I wanted to support super admin flow where we have this type or url "/admin/settings/org" and "/admin/settings/org/new" but in organization settings basic-info get mark as active because every route contains /org/settings

giancorzo: organizations/id/settings/ will show blank page... so I guess this was not ok to change it I need to think a solution for admin/settings too

NullVoxPopuli: why not change your isActive to be an exact match? and use pattern matching? 
obvs not now, but after MVP?

PR #343, opened 2018-12-14
Fix issue with a request to the current user endpoint per HoC usage

NullVoxPopuli:
Previously, the withCurrentUser HoC would cause up (and not limited) to ~ 20-30 re-renders -- ~5 per request to the current users endpoint, which there were quite a few.

Now,
there is only one request to the current user endpoint, and the component only renders twice.

Partially Resolves: https://developertown.visualstudio.com/SIL%20International%20Scriptoria/_workitems/edit/30639/
Mitigates: https://developertown.visualstudio.com/SIL%20International%20Scriptoria/_workitems/edit/28756

Note:
 - I did some refactoring as a part of this because I could just make sense of what the component structure was. Component are hierarchical, and having them all flat just kept throwing me for a loop


source/SIL.AppBuilder.Portal.Frontend/src/data/containers/with-current-user/hoc.tsx
```diff
@@ -143,52 +143,74 @@ export function withCurrentUser(opts = {}) {
...
-          this.makingRequest = false;
-          this.setState({ networkFetchComplete: true });
+          const usersFromCache = await dataStore.cache.query(cacheQuery());
+          const currentFromCache = usersFromCache[0];
+
+          if (currentFromCache) {
+            this.makingRequest = false;
+            this.setState({ currentUser: currentFromCache, isLoading: false, networkFetchComplete: true });
+          } else {
+            // we should never get here :-\
```
jonnolen: throw an error then?

source/SIL.AppBuilder.Portal.Frontend/src/data/containers/with-current-user/hoc.tsx
```diff
@@ -143,52 +143,74 @@ export function withCurrentUser(opts = {}) {
...
 
-          // this state value isn't used anywhere, but we need to trigger
-          // a re-render once the current user data has finished being consumed
-          this.makingRequest = false;
-          this.setState({ networkFetchComplete: true });
+          const usersFromCache = await dataStore.cache.query(cacheQuery());
+          const currentFromCache = usersFromCache[0];
+
+          if (currentFromCache) {
+            this.makingRequest = false;
```
jonnolen: since the final catch calls set state... should probably set `this.makingRequest = false` after that catch, right now makingRequest doesn't get reset if there is an error response. ( or it doesn't look like it does ).

source/SIL.AppBuilder.Portal.Frontend/src/data/containers/with-current-user/index.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+    return (
+      <CurrentUserContext.Provider value={{ currentUser, currentUserProps }}>
+        {children}
+      </CurrentUserContext.Provider>
+    );
+});
+
+
+export function withCurrentUserContext(InnerComponent) {
```
jonnolen: ahahhhhahahahahahahahahaaaaahhhhhhh.   nice.

jonnolen: like, i think i have a way to crack this signal-r thing now and i understand the idea of using a context to do it.

source/SIL.AppBuilder.Portal.Frontend/src/lib/i18n.tsx
```diff
@@ -1,5 +1,19 @@
 import { translate } from 'react-i18next';
+import { compose, mapProps, withProps } from 'recompose';
 
 export { InjectedTranslateProps as i18nProps } from 'react-i18next';
 
-export const withTranslations = translate('translations');
+const defaultNamespace = 'translations';
+export const withTranslations = translate(defaultNamespace);
+
+// reduce the number of props cascaded as a result of this HoC
```
jonnolen: ??

source/SIL.AppBuilder.Portal.Frontend/src/data/containers/with-current-user/hoc.tsx
```diff
@@ -71,7 +54,8 @@ export function withCurrentUser(opts = {}) {
 
         this.state = {
           currentUser: undefined, isLoading: true, error: undefined,
-          networkFetchComplete: false
+          networkFetchComplete: false,
+          fetchCurrentUser: this.fetchCurrentUser
```
jonnolen: initially i was confused why fetchCurrentUser was being put into state, but I see that we are rendering this.state into currentUserProps which is then consumable by the Provider below.  got it.

source/SIL.AppBuilder.Portal.Frontend/src/data/containers/with-current-user/index.tsx
NullVoxPopuli: yay!

source/SIL.AppBuilder.Portal.Frontend/src/lib/i18n.tsx
NullVoxPopuli: oh I was messing around with the i18n library, cause I was getting a ton of infinite renders, and I couldn't figure out what was causing them.

I'll remove my test work here

source/SIL.AppBuilder.Portal.Frontend/src/data/containers/with-current-user/hoc.tsx
NullVoxPopuli: Yeah, I needed to expose this as a result of the provider wrapping the login component.
Auth0Lock async'ly sets the auth token, but upon redirect back from the auth0 service, it was trying to get the current user before any token at been set (cause that logic wraps auth0Lock), so, now, I just let that first request happen, and then as soon as the token is set, manually re-fetch the current user.

source/SIL.AppBuilder.Portal.Frontend/src/data/containers/with-current-user/hoc.tsx
NullVoxPopuli: good _catch_

source/SIL.AppBuilder.Portal.Frontend/src/data/containers/with-current-user/hoc.tsx
NullVoxPopuli: yeah, tired me doesn't think well

source/SIL.AppBuilder.Portal.Frontend/src/data/containers/with-current-user/index.tsx
```diff
@@ -0,0 +1,60 @@
...
+  return props => {
+    return (
+      <CurrentUserContext.Consumer>
+        {(context) => <InnerComponent {...props} { ...context}/> }
+      </CurrentUserContext.Consumer>
+    );
+  };
+}
+
+export function withCurrentUser(opts = {}) {
```
NullVoxPopuli: withCurrentUserContext should be favored over withCurrentUser

source/SIL.AppBuilder.Portal.Frontend/src/data/containers/with-loader.tsx
```diff
@@ -1,16 +1,18 @@
...
+  forPage?: boolean;
 }
 
 export function withLoader<TWrappedProps>(
   isLoading: (props: TWrappedProps) => boolean,
   options: IOptions = {}
 ) {
-  const { wrapWith: Wrapper } = options;
+  const { wrapWith: Wrapper, forPage } = options;
+  const Loader = forPage ? PageLoader : RectLoader;
```
NullVoxPopuli: the PageLoaders were too large, and caused too much jitter

source/SIL.AppBuilder.Portal.Frontend/src/data/containers/with-relationship.tsx
```diff
@@ -61,29 +62,40 @@ export function withRelationships<T>(mappingFn: (props: T) => MapFnResult) {
...
+            this.isFetchingRelationships = false;
+          });
         }
       }
 
       componentDidMount() {
         this.asyncStarter();
       }
 
-      componentWillReceiveProps(newProps) {
```
NullVoxPopuli: this makes 0 sense for what this component does

PR #345, opened 2018-12-14
Implement Default Build Engine

chrisvire:
* Change `UseSilInfrastructure` to `UseDefaultBuildEngine` -- SIL should not be used in the product since someone else could deploy this and it wouldn't be SIL
* Get Default Build Engine configuration from Env Vars
* Use Default Build Engine configuration if valid

Note: I used `./run db:migrations add` to perform the column name change.  The tool went through the previous files and renamed the column.  I didn't make these changes by hand.

TerraForm defines the DEFAULT_BUILD_ENGINE_URL and DEFAULT_BUILD_ENGINE_API_ACCESS_TOKEN env vars for the api container from the deployment.

### DIFF
--- a/source/OptimaJet.DWKit.StarterApplication/Services/BuildEngine/BuildEngineSystemMonitor.cs
+++ b/source/OptimaJet.DWKit.StarterApplication/Services/BuildEngine/BuildEngineSystemMonitor.cs
@@ -41,7 +41,10 @@ namespace OptimaJet.DWKit.StarterApplication.Services.BuildEngine
         }
         private async Task SyncDatabaseTablesAsync()
         {
-            var organizations = await OrganizationRepository.GetListAsync();
+            var organizations = await OrganizationRepository
+                .Get()
+                .Where(o => o.UseDefaultBuildEngine != null && o.UseDefaultBuildEngine.Value == false)
+                .ToListAsync();
             var defaultOrganization = GetOrganizationDefaultSettings();
             if (defaultOrganization != null) { organizations.Add(defaultOrganization); }
             var statuses = await SystemStatusRepository.GetListAsync();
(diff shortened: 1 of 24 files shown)


source/OptimaJet.DWKit.StarterApplication/Migrations/20181214051403_RenameSilInfrastructureToDefaultBuildEngine.cs
```diff
@@ -0,0 +1,17 @@
+﻿using Microsoft.EntityFrameworkCore.Migrations;
+
+namespace OptimaJet.DWKit.StarterApplication.Migrations
+{
+    public partial class RenameSilInfrastructureToDefaultBuildEngine : Migration
+    {
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
```
davidmoore1: This migration doesn't appear to do anything?

chrisvire: Yeah, I wasn't sure what to do about that.  When I ran `./run db:migrations add ...` it got created AND the tool make the changes to the previous migrations.  That seemed odd, but it was helpful?  It would not have been good if we had already put this into production.

So I could remove the new files.

source/OptimaJet.DWKit.StarterApplication/Services/BuildEngine/BuildEngineSystemMonitor.cs
```diff
@@ -41,7 +41,12 @@ public void CheckBuildEngineStatus()
         }
         private async Task SyncDatabaseTablesAsync()
         {
-            var organizations = await OrganizationRepository.GetListAsync();
+            var organizations = await OrganizationRepository
+                .Get()
+                .Where(o => o.UseDefaultBuildEngine != null && o.UseDefaultBuildEngine.Value == false)
```
chrisvire: Originally I had:
`.Where(o => o.useDefaultBuildEngine.GetDefaultValue() == false)`
But I noticed in the trace logs that this statement could be translated (I assume to SQL) and had to be run locally.

NullVoxPopuli: that's a good "gotchya" for Entity Framework :+1:

source/OptimaJet.DWKit.StarterApplication/Migrations/20181214051403_RenameSilInfrastructureToDefaultBuildEngine.cs
chrisvire: Removing the files

PR #348, opened 2018-12-14
Don't use DefaultBuildEngine for sample data

chrisvire:


chrisvire:
This will be needed on the next push to master.

PR #350, opened 2018-12-14
30135 User should not be able to assign a project that does not belong to the group or organization that the project is currently in

NullVoxPopuli:
https://developertown.visualstudio.com/SIL%20International%20Scriptoria/_workitems/edit/30135

### DIFF
--- a/source/SIL.AppBuilder.Portal.Frontend/src/ui/components/inputs/user-select/with-data.tsx
+++ b/source/SIL.AppBuilder.Portal.Frontend/src/ui/components/inputs/user-select/with-data.tsx
@@ -123,7 +123,7 @@ export function withData(WrappedComponent) {
       return (users && groupMemberships && currentUsersGroupMemberships);
     }
 
-    /**
+    /*
      * Disabled if:
      * - !( the Enabled if list )
      * - User is not in the group that the project is in
@@ -133,7 +133,7 @@ export function withData(WrappedComponent) {
      * - User owns project
      * - User is org admin
      * - User is super admin
-     **/
+     */
     get isDisabled() {
       // roleInOrganization(...)
       const { currentUser } = this.props;
--- a/source/SIL.AppBuilder.Portal.Frontend/src/ui/components/inputs/user-select/with-data.tsx
+++ b/source/SIL.AppBuilder.Portal.Frontend/src/ui/components/inputs/user-select/with-data.tsx
@@ -3,12 +3,20 @@ import { compose } from 'recompose';
 import { withData as withOrbit, WithDataProps } from 'react-orbitjs';
 import { ResourceObject } from 'jsonapi-typescript';
 
-import { GroupMembershipAttributes } from '@data/models/group-membership';
-import { PageLoader as Loader } from '@ui/components/loaders';
-import { query, defaultSourceOptions, relationshipFor, USERS_TYPE, GROUP_MEMBERSHIPS_TYPE, isRelatedRecord, isRelatedTo, idFromRecordIdentity, recordIdentityFrom } from '@data';
+import {
+  query, defaultSourceOptions, relationshipFor,
+  USERS_TYPE, GROUP_MEMBERSHIPS_TYPE, isRelatedRecord, isRelatedTo, idFromRecordIdentity,
+  recordIdentityFrom,
+  OrganizationResource,
+  UserResource,
+  GroupMembershipResource
+} from '@data';
+
 import { TYPE_NAME as USER, UserAttributes } from '@data/models/user';
 import { withCurrentUser } from '@data/containers/with-current-user';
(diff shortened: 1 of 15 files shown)


source/SIL.AppBuilder.Portal.Frontend/src/ui/components/inputs/user-select/with-data.tsx
```diff
@@ -1,117 +1,169 @@
...
         restrictToGroup,
-        currentUsersGroupMemberships,
-        ...otherProps
       } = this.props;
 
-      if (!users || !groupMemberships || !currentUsersGroupMemberships) {
-        return <Loader />;
+      // remove users who _can't_ be assigned to this project
+      // due to not having organization overlap
+      const filtered = [];
```
NullVoxPopuli: refactored all of this, because the old code was pretty difficult to reason about

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/inputs/user-select/with-data.tsx
```diff
@@ -1,117 +1,169 @@
...
-        const relevantGroupMembershipsForCurrentUser = currentUsersGroupMemberships.filter(gm => {
-          return gm && validMembershipIds.includes(gm.id);
-        });
+    /*
+     * Disabled if:
+     * - !( the Enabled if list )
+     * - User is not in the group that the project is in
+     *   - consequently, user will not be in the project's organization
+     *
+     * Enabled if:
```
NullVoxPopuli: TODO: this was not functionality that existed before the refactor

source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/projects/show/display.tsx
```diff
@@ -142,7 +96,48 @@ class Display extends React.Component<IProps> {
           </div>
         </div>
 
-        <Tab menu={{ text: true }} panes={this.renderTabPanes()} className='tabs' />
+        <Tab
```
NullVoxPopuli: this didn't _need_ to happen, but it kept confusing me that I had to scroll _up_ to something that has yet to be rendered

jonnolen:
I don't see any server changes to this... this should be something enforced on the server side as well right?

NullVoxPopuli:
possibly. I don't know how important it is on the backend since groups are going away in the future.

@chrisvire / @davidmoore1 do you know if this validation is implemented?

chrisvire:
The requirement for groups for Git permissions may go away.  However, there still might be value.  That is something that we need to discuss in January.  Do not assume that they are going away.

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/inputs/user-select/with-data.tsx
```diff
@@ -1,117 +1,169 @@
...
-  groupMemberships: Array<ResourceObject<GROUP_MEMBERSHIPS_TYPE, GroupMembershipAttributes>>;
-  currentUsersGroupMemberships: Array<ResourceObject<GROUP_MEMBERSHIPS_TYPE, GroupMembershipAttributes>>;
-  usersFromCache: Array<ResourceObject<USERS_TYPE, UserAttributes>>;
-  currentUser: ResourceObject<USERS_TYPE, UserAttributes>;
+  users: UserResource[];
+  groupMemberships: GroupMembershipResource[];
+  currentUsersGroupMemberships: GroupMembershipResource[];
+  usersFromCache: UserResource[];
+  currentUser: UserResource;
   selected: Id;
```
giancorzo: just to improve readability change:
`selected: Id`
to
`selectedId: Id`

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/inputs/user-select/with-data.tsx
```diff
@@ -1,117 +1,169 @@
...
+        }
+      });
+
+      await Promise.all(promises);
+
+      // ensure the project owner is in the list
+      const isOwnerInList = filtered.find(user => user.id === selected);
+
+      if( (!isOwnerInList)) {
+        const owner = users.find(user => user.id === selected);
```
giancorzo: find can return undefined when not found. suggest to add some sort of validation here

NullVoxPopuli: I think for now, I'd rather let this error so we can figure ouwn why the selected user isn't in the users array.
I think it _shouldn't_ happen

NullVoxPopuli:
> I don't see any server changes to this... this should be something enforced on the server side as well right?

@jonnolen yes, I'll make a PBI

PR #373, opened 2018-12-21
Bugfix/user table clarity

NullVoxPopuli:


NullVoxPopuli:
so... it looks like tests that are failing locally just aren't running in ci. fun

PR #374, opened 2018-12-21
delete seed projects

NullVoxPopuli:
Projects can be created now.


NullVoxPopuli:
merging, as tests don't rely on the seed data

PR #379, opened 2019-01-02
backend docker container missing Templates

chrisvire:


jonnolen:
how likely are new directories going to be coming into the backend that need to be included in this way?  only thought is that if this happens again maybe move those files into a `resources/` directory. and just include that in the docker file.

PR #383, opened 2019-01-03
bug-30922 project files page isn't rendering / breaks react app.

jonnolen:
fix rendering to work if there are no builds, added tests.


source/SIL.AppBuilder.Portal.Frontend/src/ui/components/inputs/resource-select/index.tsx
```diff
@@ -28,6 +28,10 @@ export default class ResourceSelect<T extends ResourceObject> extends React.Comp
   render() {
     const { items, labelField, value, onChange, ...other } = this.props;
 
+    if (!items || items.length === 0){
+      return <Dropdown data-test-resource-select disabled text='No Builds Yet'/>;
```
davidmoore1: This looks like text that will appear in the app.  Should this be in the en-us.json file and pulled from there?

PR #429, opened 2019-01-10
Feedback from sil and others

NullVoxPopuli:
image
image

### DIFF
--- a/source/SIL.AppBuilder.Portal.Frontend/package.json
+++ b/source/SIL.AppBuilder.Portal.Frontend/package.json
@@ -69,6 +69,7 @@
     "react-router": "^4.3.1",
     "react-router-dom": "^4.3.1",
     "react-timezone": "^2.1.0",
+    "react-toastify": "^4.5.2",
     "recompose": "^0.30.0",
     "redux": "^4.0.0",
     "redux-actions": "^2.6.3",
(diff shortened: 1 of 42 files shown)


chrisvire:
What should I be reviewing?

NullVoxPopuli:
@chrisvire I guess the screenshots?

NullVoxPopuli:
latest status styles:
image

NullVoxPopuli:
Latest toast style: 
image

chrisvire:
Which aspects of the screen shot?  It would be nice to have a bullet list of the things you changed.

BTW, the green color on the toast is obnoxious.
The gray oval look less like buttons.  Maybe Josh should give some input?
Private icon is fine.

Anything else I should be noticing?

NullVoxPopuli:
image


what about that?


The list:
 - new tasks list
 - private projects no longer have a blue background
 - new toasts

NullVoxPopuli:
image

NullVoxPopuli:
I added fixes to 30930 and 30370 as well (they were small)
all that's remaining is fixing the tests for product selection, cause they were tied to order...
just need to update them to select products by name instead of by index.

chrisvire:
LGTM (after tests pass)

PR #450, opened 2019-01-18
Sync SuperUser Role to DWKit Admins Role

chrisvire:


NullVoxPopuli:
yeah, I think some easier tests could be a side-effect of a broader refactor w/ auto-matter (eventually) now that we kiiinda have a sync pattern goin'.

but anyway, LGTM

PR #491, opened 2019-01-23
Feature/implement language picker and search

NullVoxPopuli:
Good news, even though this PR is huge, some of it is yarn.lock shenanigans:
image

 - implement code splitting to better handle very large files
   - language list is > 3MB
 - localize the alltags.json file's country codes, see: [this doc](https://docs.google.com/document/d/1nKziDjjAZEO3vb05442urQ4i86-_ksmk4-pfSEdKP9o/edit#heading=h.u2v46jj363v)
 - ux:
   - input takes free-form text
   - field turns yellow if the text does not exist as tag in the alltags list
   - fully keyboard accessible
 - performance:
   - only 5 results are rendered at once
     - this allows search results to appear for any amount of entered characters

image
image


chrisvire:
Answers:
>* backend validation? (new projects require the language be from alltags.json?)
No backend validation.  These are just suggestions to help them use a good code.  There are cases where there isn't a registered code yet.  We might do something where we notify a Org Admin of unrecognized code being used.
>* are there duplicate tags?
There shouldn't be duplicates in the canonical tags ("tag" property).  There could be duplicates in the alternate tags ("tags" property). I am checking with @mhosken.

chrisvire:
I talked with one of our designers (@merchako).  This particular UI is important to him that we share it across our products and will be spending time over the next 6 months designing it.  With that said, he said to do what is easiest and be able to replace later.  Technically, what is easiest is what is there. :-)  But he does have feedback.  I will attempt to summarize.  Please give your feedback.

We can do the dropdown, but there is too much information.  It would be better if:
1. Have LocalName in upper left in larger text (no label) -- this is the language name in the local script
2. Have localized language name (from LDML) below it in regular size -- this is the language name localized into the language of the current UI
3. Have the canonical code on the right (can have "code" label)
4. on hover (yeah, just desktop), show a panel next to the dropdown with the full details which would include other tags, regions, and names.  It could repeat the primary information so that it was a full profile (this could be useful when display info about the language code in other parts of the UI). [If you have a suggestion for mobile, I am open to hear it).

NullVoxPopuli:
that feedback sounds good, but my only issue is with #4, where hovering goes against a11y patterns, we could have a small link ('`more info`'?) that shows the same ui / details. This would also allow it to work on mobile.

chrisvire:
OK.  I am fine with that.  Also, when highlighting, if there highlighted details that are only in the details, then you could highlight the "more info".  Would clicking/tapping on more info just expand the current entry?

NullVoxPopuli:
>  Also, when highlighting, if there highlighted details that are only in the details, then you could highlight  the "more info". Would clicking/tapping on more info just expand the current entry?

it's misleading UX to search over things that are not visible in the default view, and I think it adds some gnarly complexity in order to _make_ those things viewable upon rendering of search results.


> How about just "More" instead of "More Info"?

sounds good

merchako:
Great point about a11y, Preston.
I like the compromise of using *more *or *more info* to let the user
explore more.

I have an alternate compromise that should:
(1) address Preston's "search over invisible things" concern,
(2) be easier to implement if you do in lieu of *More*, and
(3) be compatible with *More* if you decide to do both.

*Here it is:*

Row 1 left: autonym of language (large text)
Row 2 left: name of language in interface language (small text) (if
different from autonym)

Row 1 right: BCP-47 code of language (large text)
Row 2 right: a label saying "code", "language code", or even "BCP-47 code"
(small text)

Row 3 left: an excerpt from other searchable fields showing some places
where the search query appears in them, if applicable (just like in Google
search results) (small text)
Row 3 right: *More *link (optional) (small text)

Each item in the dropdown list should be either 2 lines or 3 lines. If
there's no need to show the excerpts of the 3rd row, hide the 3rd row.

Let me know if you want/need a mockup.

—Alex Mercado, Digital Product Designer, SIL International

NullVoxPopuli:
re: the 3rd row, if you already have a dynamically shown based on if it contains a match, why do we need a more button for other things? the search already includes the language? why would you need to see more info on it?

merchako:
That’s a great point, which is why I mark that element as optional. For iterative development, I would guess it’d be easiest to leave it out for now and not give it design or development time.

—Alex Mercado

NullVoxPopuli:
ok, so... atm, I have this: 

image

The type interface for this is:
```ts
interface ILanguageInfo {
  full: string;
  iana: string;
  iso639_3: string;
  localname?: string;
  name: string;
  names?: string[];
  region: string;
  regionname: string;
  regions?: string;
  sldr: boolean;
  tag: string;
  tags?: string[];
}
```

obviously, a number of these can be combined for non-default fields.
default fields are:

```
{localname || name}                   {tag}
{localizedName(tag)}                  CODE         

{additionalMatch}
```

where, `localizedName(tag)` is the value of `ldml.localeDisplayNames.languages[tag]`

So, for the remaining fields, 
```ts
  full: string;
  iana: string;
  iso639_3: string;
  names?: string[];
  region: string;
  regionname: string;
  regions?: string;
  sldr: boolean;
  tags?: string[];
```

I figure `sldr` isn't something that's search-overable, so, I'll ignore that one (unless we need that for some reason).

Currently, everything is search-overable, (except the sldr field), 
but, as far as the additional data / what appears in the 3rd row, I can (maybe) combine these:

`region`, `regions`
`iana`, `name`, `names`, (does `localname` matter?)
`tag`, `tags`, `iso639_3`, `full`

that leaves: 
`regionname`

So that means I need 4 L10n'd labels(?):
Region(s)
Name(s)
Tag(s)
Region Name

the LDML Data doesn't seem to have anything about this (label-wise), so I assume that the field _label_ we can manage in our own translation files..

And just to clarify, the way to L10n the rest of the data from the LDML data:
```
  full: string;     => no translation
  iana: string;     => no translation
  iso639_3: string;     => no translation
  names?: string[];      => no translation
  region: string;       => ldml.localeDisplayNames.territories[region]
  regionname: string; => no translation
  regions?: string;   => regions.split(' ').map(r => ldml.localeDisplayNames.territories[r]);
  sldr: boolean;      => ignore
  tags?: string[];       => no translation
```

is this correct?

chrisvire:
> ok, so... atm, I have this:
> [snip]
> is this correct?

Yes, I believe that is correct.

chrisvire:
> re: the 3rd row, if you already have a dynamically shown based on if it contains a match, why do we need a more button for other things? the search already includes the language? why would you need to see more info on it?

If there are two that are similar, I could speculate you might need the full details to see the difference.  I don't have full examples.  We could defer to later.

merchako:
That’s the general idea. Any additional comments I would make are nit-picking, so for the sake of speed, you can proceed as you wish.

(1) When you use labels, make sure they’re human-friendly, which `LOCALE-PICKER.REGIONS`, for example, is not. 
(2) Maybe you don’t need to label the _context_ that you show in the last row, especially since those labels aren’t meaningful anyway. That would help to keep the size of each list item small by eliminating a row. This is especially important if you’re showing contextual matches from multiple fields, which would be desirable.
(3) Make sure to set `.context-selector { text-overflow: clip }` to keep text from rendering outside of a list item.
(4) To improve visual hierarchy, make the text size for _autonym_ and _language code_ bigger than that of _context_. I don’t have a live demo of this, so I can’t say whether you should make the former bigger or the latter smaller.
(5) Make sure to expands the abbreviations in `LOCALE-PICKER.REGIONS` (which I think Chris Hubbard mentioned that you would already do).
(6) If possible, the dropdown should hang flush from the bottom of the query field, similar to how search suggestions behave on google.com.

NullVoxPopuli:
> (1) When you use labels, make sure they’re human-friendly, which `LOCALE-PICKER.REGIONS`, for example, is not.

That was just a placeholder, here is what it looks like now:

image



> (2) Maybe you don’t need to label the _context_ that you show in the last row, especially since those labels aren’t meaningful anyway. That would help to keep the size of each list item small by eliminating a row. This is especially important if you’re showing contextual matches from multiple fields, which would be desirable.

:man_shrugging: 

> (3) Make sure to set `.context-selector { text-overflow: clip }` to keep text from rendering outside of a list item.

why? that cuts off words mid... word

> (4) To improve visual hierarchy, make the text size for _autonym_ and _language code_ bigger than that of _context_. I don’t have a live demo of this, so I can’t say whether you should make the former bigger or the latter smaller.

I don't actually know what autonym and context are, so... yeah. I have `name`,` `localized name`, `tag` (code), and additional matches below, which could be anything, and happen anywhere within the bigger blocks of text. so for bigger blocks of text, the viewport needs to be centered on the highlight, rather than rendering from the beginning.

> (5) Make sure to expands the abbreviations in `LOCALE-PICKER.REGIONS` (which I think Chris Hubbard mentioned that you would already do).

this was just a placeholder until I filled in the translations file

> (6) If possible, the dropdown should hang flush from the bottom of the query field, similar to how search suggestions behave on google.com.

yeah, good catch, I copied all the styles from the demo site.
(already applied this change in the above screenshot)

merchako:
(1) 👍🏼
(2) It'd be nice to see what this looks like. I assume it's just one or two lines of CSS to see it without the label?
(3) `text-overflow: ellipsis` is also fine
(4) autonym = name of the language in the language, which I assume is localized name. What I mean is that `localized language name` and `code` should be visually bigger than "additional matches".
(5) 👍🏼
(6) 👍🏼

NullVoxPopuli:
@merchako 
image
image
image

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/inputs/locale-input/index.tsx
```diff
@@ -0,0 +1,45 @@
...
+export default class extends React.Component<IProps, IState> {
+  loading = true;
+  state: IState = {};
+
+  componentDidMount() {
+    this.fetchLanguageData();
+  }
+
+  async fetchLanguageData() {
+    const response = await fetch('/assets/language/alltags.json');
```
NullVoxPopuli: this is a 3MB file, so it's not bundled with the code

source/SIL.AppBuilder.Portal.Frontend/src/data/store.ts
```diff
@@ -1,39 +1,11 @@
-import { Bucket } from '@orbit/core';
```
NullVoxPopuli: this stuff was all pulled out into a builtin strategy to react-orbitjs, here: https://github.com/developertown/react-orbitjs/blob/master/src/strategies/pessimistic-with-remote-ids.ts

source/SIL.AppBuilder.Portal.Frontend/src/lib/debug.tsx
```diff
@@ -23,6 +23,16 @@ export const logProps = (label: string) => (WrappedComponent) => {
   };
 };
 
+export function measureTime(label: string, fn: (...args: any[]) => void) {
```
NullVoxPopuli: debug tool I used for seeing how long certain functions took to run

source/SIL.AppBuilder.Portal.Frontend/src/lib/i18n.tsx
```diff
@@ -1,6 +1,17 @@
...
 export { TransProps as i18nProps } from 'react-i18next';
 
 const defaultNamespace = 'translations';
-export const withTranslations = translate(defaultNamespace);
+const ldml = 'ldml';
+
+export const withTranslations = withTranslation(defaultNamespace);
+export const withLdml = withTranslation(ldml);
+
+export function useLdml() {
```
NullVoxPopuli: hooks!

source/SIL.AppBuilder.Portal.Frontend/src/translations/fetch-l10n.tsx
```diff
@@ -0,0 +1,38 @@
...
+  const { i18n } = useTranslation();
+  const { language } = i18n;
+
+  if (i18n.hasResourceBundle(language, namespace)) {
+    return children;
+  }
+
+  return (
+    <React.Suspense fallback={<PageLoader />}>
+      <LocaleLoader />
```
NullVoxPopuli: when LocaleLoader is fetching data, a promise is `throw`n which `Suspense` catches and then renders the fallback component. Once the promise resolves, the children are rendered.

source/SIL.AppBuilder.Portal.Frontend/src/translations/index.tsx
```diff
@@ -1,10 +1,8 @@
 import i18n from 'i18next';
-import * as ICU from 'i18next-icu';
-import { reactI18nextModule } from 'react-i18next';
-import * as LanguageDetector from 'i18next-browser-languagedetector';
-
-// import * as Backend from 'i18next-xhr-backend';
-// import { reactI18nextModule } from 'react-i18next';
+import ICU from 'i18next-icu';
```
NullVoxPopuli: result of some upgrading as I tried to use i18next to do things for me.
supposedly it supports async/remote translations, but the lack of fetch/xhr invocation told me otherwise.

that's why `fetch-l10n.tsx` exists now. :(

source/SIL.AppBuilder.Portal.Frontend/src/ui/application.tsx
```diff
@@ -46,27 +48,29 @@ export default class Application extends React.Component<IProps> {
...
-                      <Component />
-                    </ScrollToTop>
-                    <DebugInfo />
-                  </>
-                </Router>
-              </ReduxProvider>
-            </SocketManager>
-          </CurrentUserProvider>
-        </APIProvider>
+        <L10nLoader>
```
NullVoxPopuli: only `<L10nLoader>` was added

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/inputs/date-range/styles.scss
```diff
@@ -7,18 +7,17 @@
...
-    position: absolute !important;
-    right: 0;
-  }
-}
-
 .date-range-input-with-label {
   svg {
     pointer-events: none;
   }
+  .input {
```
NullVoxPopuli: I had to adjust the filters on the project directory screen to use css-grid, which resulted in some dynamic positioning at certain screen sizes. I noticed that the svg/icons were not always correctly positioned, so this had to change.

for reference: this is the filter list:
image
image
image

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/inputs/locale-input/-page.ts
```diff
@@ -0,0 +1,100 @@
...
+class LocaleInput extends Interactor {
+  static defaultScope = '[data-test-locale-input-container]';
+
+  async searchAndSelect(searchTerm: string) {
+    await when(() => assert(this.isPresent, `\nexpected language picker to be visible\n`));
+
+    await this.input.fill(searchTerm);
+    // bigtest blur/focus don't work?
+    Simulate.blur(this.input.$root);
+    Simulate.focus(this.input.$root);
```
NullVoxPopuli: The react-autosuggest input activates strangely in a programmatically driven environment, not sure what that's about -- whether it'd be a bug with them, or what, but I took this technique from their own tests.

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/inputs/locale-input/-utils/find-additional-matches.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+interface IAdditionalSearchResult {
+  match?: MatchResult;
+  additionalMatches: MatchResult[];
+}
+
+const translationMap = {
+  full: 'tag',
+  iana: 'tag',
+  ['iso639_3']: 'tag',
```
NullVoxPopuli: this gets around a linting rule -.-

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/inputs/locale-input/-utils/find-additional-matches.tsx
```diff
@@ -0,0 +1,73 @@
...
+  if (has('localname') || has('name') || has('tag')) {
+    return result;
+  }
+
+  Object.keys(suggestion).forEach((key) => {
+    if (keysToSkip.includes(key)) {
+      return;
+    }
+    let rawValue = suggestion[key];
+    // TODO: localize separator, and values
```
NullVoxPopuli: punted for now, due to time constraints and already blown up scope

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/inputs/locale-input/-utils/localize.ts
```diff
@@ -0,0 +1,41 @@
...
+      //  - name: string;
+      //  - names?: string[];
+      //  - regionname: string;
+      //  - sldr: boolean;
+      //  - tag: string;
+      //  - tags?: string[];
+
+      // These are the only fields that need localization
+      region: info.region && tryLocalize('territories', info.region),
+      regions:
```
NullVoxPopuli: regions is a string, but _is_ space-delimited.

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/inputs/locale-input/field.tsx
```diff
@@ -0,0 +1,142 @@
...
+  getSuggestions: (value: string) => ILanguageInfo[];
+  findLanguageCode: (value: string) => string;
+  findLanguageInfo: (value: string) => ILanguageInfo;
+
+  constructor(props) {
+    super(props);
+
+    assert(`The Locale Picker Field needs a data set. The passed data set is empty.`, props.data);
+
+    // TODO: re-calc this if the user changes their language
```
NullVoxPopuli: I have a plan to convert this to react-hooks later, which will allow me to memoize a specific property within the function, which will allow updates when the currently selected language changes.

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/inputs/organization-select/display.tsx
```diff
@@ -36,6 +37,7 @@ class Display extends React.Component<IProps> {
       options: organizationOptions,
       onChange: this.onChange,
       value: defaultValue,
+      icon: <ArrowDropDownIcon />,
```
NullVoxPopuli: this icon not existing isn't directly related to this set of changes, but the icon was missing, and I figured I might as well fix it while I'm in here

source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/admin/index.tsx
```diff
@@ -4,17 +4,16 @@ import { Switch, Route } from 'react-router-dom';
...
 import AdminSettingsRoute, { pathName as settingsPathName } from './settings';
 
-export const pathName = '/admin';
-
 class AdminRoute extends React.Component {
   render() {
     return (
       <Switch>
-        <Route exact path={pathName} component={InviteOrganization} />
+        <Route exact path={adminPath} component={InviteOrganization} />
```
NullVoxPopuli: all path changes were due to experimenting with lazy-loading / code-splitting on routes early on in this PR when I was blocked from actually working on the language picker.

I learned some things about code splitting, which I don't think is well known: https://twitter.com/nullvoxpopuli/status/1093512581896114176
basically, code-splitting isn't always beneficial. testing to see if it has any benefit is a good thing to do.

source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/invitations/index.tsx
```diff
@@ -1,8 +1,5 @@
 import * as React from 'react';
 import { Switch, Route } from 'react-router-dom';
-import { compose } from 'recompose';
```
NullVoxPopuli: these weren't used

source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/paths.ts
```diff
@@ -0,0 +1,13 @@
+// Top level routes
+export const rootPath = '/';
```
NullVoxPopuli: I think the next iteration of this would look something like this:

```ts
export const routeFor(routeSegment) {
  const { name, path } = routeSegment;

  if (name) {
    return name;
  }

  if (typeof path === 'function') {
    throw new Error('path is a function and expected a name property to be present');
  }

  return path;
}

export const paths = {
  root: {
    path: '/',
  },
  login: { path: '/login', },
  projects: {
    path: '/projects',
    new: { path: /projects/new', },
    show: { 
      name: '/projects/:id',
      path(id: string) { return `/projects/${id}`; },
    },
    edit: { 
      name: '/projects/:id/edit',
      path(id: string) { return `/projects/${id}/edit`; } 
    },
  }
}
```

or something like that. idk. #ReactRouterProblems

source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/project-directory/list/__tests__/filtering-by-organization-acceptance-test.ts
```diff
@@ -40,6 +40,11 @@ describe('Acceptance | Project Directory | Filtering | By Organization', () => {
         expect(req.headers.Organization).to.equal(undefined);
       });
     });
+
+    server.get('/assets/language/alltags.json').intercept((req, res) => {
+      res.status(200);
+      res.json([]);
```
NullVoxPopuli: the actual file has about 9k entries

source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/project-directory/list/__tests__/filtering-by-organization-acceptance-test.ts
```diff
@@ -78,8 +83,10 @@ describe('Acceptance | Project Directory | Filtering | By Organization', () => {
       let rows;
 
       beforeEach(async function() {
-        requestCount = 0;
         await when(() => page.table.rows().length === 3);
+        // page has finished loading ^
```
NullVoxPopuli: this test was wrong and was passing due to async race-condition magic / sadness

source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/project-directory/list/filters/index.tsx
```diff
@@ -107,34 +123,35 @@ class Filter extends React.Component<IProps, IState> {
...
-          </div>
-          <div className='input flex-row align-items-center m-l-md m-r-md'>
-            <OrganizationSelect
-              className='w-100'
-              onChange={this.handleOrganizationChange}
-              organizations={organizations}
-              defaultValue={selectedOrganization}
-            />
-          </div>
+      <div className='grid-container m-b-lg'>
```
NullVoxPopuli: switched from flex col|row layout to grid

merchako:
> @merchako
> image
> image
> image

If I had my way, I'd go with:
(1) No label on the last row (your second screenshot)
(2) Smaller text for the last row.
Granted, we're just talking about my opinion now.

In other news, the label (middle left) underneath the autonym (language as it calls itself) is wrong. You should be showing the name of the language as the interface language calls it. Instead, you're showing the language code.

source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/users/list/__tests__/invite-user-acceptance-test.tsx
```diff
@@ -206,17 +209,19 @@ describe('Acceptance | User list | Invite User', () => {
           });
 
           describe('clears error on close', () => {
+            let modal;
             beforeEach(async () => {
-              await page.inviteUser();
+              modal = new InviteUserModalInteractor('[data-test-users-invite-user-modal]');
```
NullVoxPopuli: this was an intermittently failing test, and I had very little trust in the OG interactor, so I `new`d a new one.
also, I think this test may be testing something that can't actually happen, I don't remember at this point. :-\

source/SIL.AppBuilder.Portal.Frontend/src/ui/styles/components/locale-auto-suggest.scss
```diff
@@ -0,0 +1,110 @@
+.phrase-with-highlight {
+  .highlight {
+    background: yellow;
+  }
+}
+
+
+.locale-input__has-match,
```
NullVoxPopuli: a lot of this was copied from react-auto-suggest's demo. it was tweaked to allow for sizing / positioning, etc changes

NullVoxPopuli:
> If I had my way, I'd go with:
> (1) No label on the last row (your second screenshot)
> (2) Smaller text for the last row.
> Granted, we're just talking about my opinion now.

for the sake of getting this done, I'm going to punt this for now.

> In other news, the label (middle left) underneath the autonym (language as it calls itself) is wrong. You should be showing the name of the language as the interface language calls it. Instead, you're showing the language code.

I looked into this, and it seems, that for `zzj-Hani` there is no localized name for English, using the ldml stuff from the SIL API anyway. 
However, that second entry now correctly says Cantonese:
image

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/inputs/locale-input/-utils/find-additional-matches.tsx
abillingsley: do we have a story somewhere in the backlog to revisit this at the appropriate time?

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/inputs/locale-input/-utils/helpers.tsx
```diff
@@ -0,0 +1,158 @@
+import * as React from 'react';
+
+export const hasSearchTerm = (searchTerm: string) => {
+  let lowerSearch = (searchTerm || '').trim().toLocaleLowerCase();
```
abillingsley: `searchTerm` is defined as a string but treated as if it is null or undefined here.

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/inputs/locale-input/-utils/helpers.tsx
```diff
@@ -0,0 +1,158 @@
...
+
+  return (property: string) => {
+    try {
+      return (property || '').toLocaleLowerCase().includes(lowerSearch);
```
abillingsley: `property` is defined as a string but treated as if it is null or undefined here.

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/inputs/locale-input/-utils/helpers.tsx
```diff
@@ -0,0 +1,158 @@
...
+export function findLanguageCode(data: ILanguageInfo[]) {
+  return (value: string) => {
+    const info = data.find((lang) => lang.tag === value);
+
+    return (info && info.tag) || '';
+  };
+}
+
+export function findLanguageInfo(data: ILanguageInfo[]) {
+  return (value: string) => {
```
abillingsley: as a one liner?
```
return (value: string) => data.find((lang) => lang.tag === value);
```

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/inputs/locale-input/-utils/helpers.tsx
```diff
@@ -0,0 +1,158 @@
...
+//
+// this is a straight port of the C# code
+//
+/// Sorting the languages for display is tricky: we want the most relevant languages at the
+/// top of the list, so we can't simply sort alphabetically by language name or by language tag,
+/// but need to take both items into account together with the current search string.  Ordering
+/// by relevance is clearly impossible since we'd have to read the user's mind and apply that
+/// knowledge to the data.  But the heuristics we use here may be better than nothing...
+export function sortComparer(searchTerm?: string) {
+  const lowerSearch = `${searchTerm}`.toLocaleLowerCase();
```
abillingsley: if `searchTerm` is undefined `lowerSearch` will equal `"undefined"` is this ok?

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/inputs/locale-input/field.tsx
```diff
@@ -0,0 +1,141 @@
...
+  constructor(props) {
+    super(props);
+
+    assert(`The Locale Picker Field needs a data set. The passed data set is empty.`, props.data);
+
+    this.getSuggestions = getSuggestions(props.data).bind(this);
+    this.findLanguageCode = findLanguageCode(props.data);
+    this.findLanguageInfo = findLanguageInfo(props.data);
+
+    const value = this.findLanguageCode(props.value) || '';
```
abillingsley: what would happen if `findLanguageCode` is not found and `value` is empty? should we throw here or will this behave properly?

abillingsley: also should `findLanguageCode` encapsulate the logic around returning empty when not found?

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/route-listener/index.tsx
```diff
@@ -29,8 +29,8 @@ export const RouteListener = compose(
 
         // TODO: if in some debug mode, log transitions and such
         //      (handy for debugging tests)
-        // console.log(`The current URL is ${location.pathname}${location.search}${location.hash}`);
-        // console.log(`The last navigation action was ${action}`);
+        console.log(`The current URL is ${location.pathname}${location.search}${location.hash}`);
+        console.log(`The last navigation action was ${action}`);
```
abillingsley: intended or just debugging?

source/SIL.AppBuilder.Portal.Frontend/src/ui/styles/semantic-ui-overrides.scss
```diff
@@ -88,6 +88,14 @@ Inputs
   background-color: $checkbox-background-color;
 }
 
+.input {
+  position: relative;
+  svg, i {
+    position: absolute !important;
```
abillingsley: a couple more `important!`s here an in the nex file.  necessary?

source/SIL.AppBuilder.Portal.Frontend/scripts/refresh-languages.sh
```diff
@@ -0,0 +1,52 @@
...
+# Usage:
+#
+#   ./scripts/refresh-languages.sh
+#   yarn webpack:build or yarn start:dev
+#
+# NOTE: the CWD/PWD should be the frontend folder. not scripts.
+set -e
+
+
+url="https://raw.githubusercontent.com/silnrsi/sldr/master/extras/alltags.json"
```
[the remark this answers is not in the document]
NullVoxPopuli: yup, that's intentional / what the other SIL projects do

source/SIL.AppBuilder.Portal.Frontend/scripts/refresh-languages.sh
```diff
@@ -0,0 +1,52 @@
...
+targetPath="${folder}alltags.json"
+
+mkdir -p $folder
+
+echo "downloading all supported languages..."
+wget --output-document $targetPath $url
+echo "downloaded all supported languages"
+
+echo "downloading transaltion support for the language data..."
+npm install -g fast-xml-parser
```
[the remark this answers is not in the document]
NullVoxPopuli: idk, I had conflicting feelings about that, because it's not used anywhere else in the project.

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/inputs/date-range/styles.scss
```diff
@@ -7,18 +7,17 @@
...
+    position: relative;
+    svg {
+      position: absolute !important;
```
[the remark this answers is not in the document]
NullVoxPopuli: it's hacking around SemanticUI. They don't know how to write CSS, and they, themselves, use important everywhere :cry: :sob: :sob:

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/inputs/locale-input/-utils/find-additional-matches.tsx
NullVoxPopuli: we do now! https://developertown.visualstudio.com/SIL%20International%20Scriptoria/_workitems/edit/33104

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/inputs/locale-input/-utils/helpers.tsx
NullVoxPopuli: yeah,
I've changed this to `searchTerm?: string` in the upcoming commit.

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/inputs/locale-input/-utils/helpers.tsx
NullVoxPopuli: yeah, it could be. I've been hesitant to go overboard with one liners, because it makes debugging annoying (gotta re-convert to a `{}` function, add `debugger` (or a breakpoint), then convert back, etc.
idk

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/inputs/locale-input/-utils/helpers.tsx
NullVoxPopuli: nope! thanks!

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/inputs/locale-input/field.tsx
NullVoxPopuli: findLanguageCode just does `Array#find`, so it'd return undefined if there is no result. if value is empty, then undefined will also be returned, as every value it searches over isn't `undefined`. (this will behave properly)

I don't think findLanguageCode should encapsulate the `|| ''` because the `|| ''` is specifically to force the input in the template of this component to be a "controlled" component. In the event that an input's value is undefined, the input is "uncontrolled", and React yells at you for converting it from uncontrolled to controlled at runtime. It's silly, cause `undefined` _can't_ be a value of an input field, and I think the `|| ''` should actually be a JSX concern :-\

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/route-listener/index.tsx
NullVoxPopuli: thanks for catching this, needs to be recommented out

source/SIL.AppBuilder.Portal.Frontend/src/ui/styles/semantic-ui-overrides.scss
NullVoxPopuli: yeah :(

Never using SemanticUI again.

Had I known about ShadowDom before this project, I would have mounted all the SemanticUI requiring things into a shadow dom, and use Material for everything else (or any other css framework that knows how to write css)

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/inputs/locale-input/-utils/helpers.tsx
abillingsley: I hear ya. definitely not a deal breaker by any means

PR #493, opened 2019-01-23
Change StatusUpdateService start from Hangfire task to IHostedService

chrisvire:


chrisvire:
I would like to handle lost connections too.

NullVoxPopuli:
my preference would be to have these "notifications" moved into app land, where after a record-of-interest is created, you push a db notification (create, table-name, id, etc) into some hangfire queue, and have a hangfire job watch the queue and do whatever, if needed

chrisvire:
> my preference would be to have these "notifications" moved into app land, where after a record-of-interest is created, you push a db notification (create, table-name, id, etc) into some hangfire queue, and have a hangfire job watch the queue and do whatever, if needed

We attempted to use the `AppDbContext::SaveChanges{Async}`, however we didn't get the Id field on insert which made it not as useful.  If we want to limit the notifications to modifications, then maybe that would be OK.  However, one use case is sitting on the UserTasks page.  Without the new row's Id, would you have to re-execute the whole query (not just query that one object that was created)?

NullVoxPopuli:
> We attempted to use the AppDbContext::SaveChanges{Async}, however we didn't get the Id field on insert which made it not as useful.

I was just saying to create the notification in the (for example) ProjectService.cs at the end of `Create` or something like that -- since all CRUD actions have to go through the service, we could fake the existence of lifecycle hooks by just adding to the service -- which would allow access to the id. :)

> Without the new row's Id, would you have to re-execute the whole query (not just query that one object that was created)?

as long as we aren't doing this multiple times a second, I don't think re-fetching data is bad if it can't be avoided

chrisvire:
> since all CRUD actions have to go through the service

Those services are tied to frontend calls.  Data is also modified through `IJobRepository<>` by the backend jobs.  These are more likely the changes we want to be notified about!

NullVoxPopuli:
Is there a reason the IJobRepository stuff isn't using the services?
the service-repo layer is supposed to be the only layer we interact with the database through

chrisvire:
> Is there a reason the IJobRepository stuff isn't using the services?

The Services are tied to `IJsonApiContext` which is API calls

PR #495, opened 2019-01-23
Expose hangfire dashboard to public

NullVoxPopuli:
Currently always allows access.

I think I saw somewhere that we could do basic auth, but I lost that link -- anyway that might be an option if we don't want to integrate with our existing auth.  Big downside is that http auth is insecure, and we'd only have one username/password


chrisvire:
Please merge and deploy to dev

PR #504, opened 2019-01-24
Live Data Updates

NullVoxPopuli:


NullVoxPopuli:
blocked by: https://github.com/sillsdev/appbuilder-portal/pull/555

source/SIL.AppBuilder.Portal.Frontend/src/data/orbitjs-operations-support/serialize-to-api.ts
```diff
@@ -0,0 +1,187 @@
+import {
```
NullVoxPopuli: this stuff and all related stuff will (eventually) get pulled directly into orbit.js,

NullVoxPopuli:
Work on integrating a bunch of this into orbit is happening over here: https://github.com/orbitjs/orbit/pull/597

source/OptimaJet.DWKit.StarterApplication/BackendServicesExtension.cs
```diff
@@ -55,7 +56,7 @@ public static IServiceCollection AddApiServices(this IServiceCollection services
             services.AddScoped<IEntityRepository<Notification>, NotificationRepository>();
             services.AddScoped<IEntityRepository<Product,Guid>, ProductRepository>();
             services.AddScoped<IEntityRepository<OrganizationStore>, OrganizationStoreRepository>();
-
+            services.AddScoped(typeof(IUpdateService<Project, int>), typeof(ProjectService));
```
muchman: any reason this signature is different than the others?
```suggestion
            services.AddScoped<IUpdateService<Project, int>>, ProjectService>();
```

muchman: for my own ocd reasons I would just standardize all of these to one syntax

NullVoxPopuli: there isn't. probably got caught in using typeof with generics, and just didn't switch it back. thanks!

source/OptimaJet.DWKit.StarterApplication/EventDispatcher/EnitityEventHandler/BaseHookNotifier.cs
```diff
@@ -11,70 +21,186 @@
...
 
-            dynamic message = new ExpandoObject();
-            message.table = tableName;
-            message.id = id;
-            message.operation = operation;
 
-            var updateGroups = GetUpdateGroups(tableName, id);
-            var json = JsonConvert.SerializeObject(message);
-            SendStatusUpdateAsync(updateGroups, json).Wait();
+            dynamic idInEntityType; // Guid | int | string
```
muchman: personally I would use object instead of dynamic here.  I dont think you are gaining anything by using a dynamic since DbSet().Find() takes an object[] as its params.  dynamic will allow you do things that cause runtime errors, like access properties that dont exist.

source/SIL.AppBuilder.Portal.Backend.Tests/Acceptance/EventDispatcher/EntityHandler/ProjectHookHandlerTest.cs
```diff
@@ -72,62 +72,56 @@ public void Notify_Project_Insert()
         {
             BuildTestData();
 
-            var handler = _fixture.GetService<IEntityHookHandler<Project>>();
-            var hubContext = _fixture.GetService<IHubContext<ScriptoriaHub>>();
-            var mockScriptoriaHub = Mock.Get(hubContext);
-            mockScriptoriaHub.Reset();
+            var handler = _fixture.GetService<IEntityHookHandler<Project, int>>();
```
muchman: Not sure how helpful it would be, but this kind of pattern with the hubContext seems to be repeated a lot.  Maybe refactor this into some shared function?

PR #514, opened 2019-02-02
Generate notifications during Repository changes

chrisvire:


source/OptimaJet.DWKit.StarterApplication/Repositories/Background/JobRepository.cs
```diff
@@ -27,17 +29,24 @@ public class JobRepository<TEntity, TId>
...
             this.dbSet = contextResolver.GetDbSet<TEntity>();
+            StatusUpdateService = statusUpdateService;
         }
 
         public virtual async Task<TEntity> CreateAsync(TEntity entity)
         {
             dbSet.Add(entity);
 
             var retval = await dbContext.SaveChangesAsync();
+            StatusUpdateService.OnInsert(entity as IStatusUpdate);
```
NullVoxPopuli: I think status update service may be too specific of a name.

generally, what we want to capture any crud operation.
So, maybe I'd have something named: `EntityHooksService` with methods, like, `didCreate`, `didUpdate`, `didDestroy` -- past tense-ily named methods because the action already happened. 

Also, if we generally take any entity, instead of an `IStatusUpdate` and just define an entity as always defining a method that does Id.ToString(), then we can use this more a lot more than the current implementation.

But, for each of the hook methods, we'll probably want to pass a generic type as well, so if needed, we can have the EntityHooksService spin off background jobs based off the type of record (which would be beneficial of something is updating often, and would allow us to do things like keep the latest update, and drop all the yet-to-be-run notifications)

source/OptimaJet.DWKit.StarterApplication/Repositories/BaseRepository.cs
```diff
@@ -40,24 +28,52 @@ public class BaseRepository<TEntity, TId> : DefaultEntityRepository<TEntity, TId
...
+            var retval = await base.CreateAsync(entity);
+            statusUpdateService.OnInsert(retval as IStatusUpdate);
+            return retval;
+        }
+
+        public override async Task<bool> DeleteAsync(TId id)
+        {
+            var entity = await GetAsync(id);
+            var retval = await base.DeleteAsync(id);
+            if (retval)
```
NullVoxPopuli: this event could still be published without a `retval`, we have the id and we should have a T, where T is some Entity type.

what situations does deleteAsync not return something? if it doesn't we also have the original entity from getasync.

I just think it's weird to optionally use the delete hook

source/OptimaJet.DWKit.StarterApplication/Repositories/OrganizationRepository.cs
```diff
@@ -18,11 +18,12 @@ public class OrganizationRepository : BaseRepository<Organization>
...
             IJsonApiContext jsonApiContext,
             CurrentUserRepository currentUserRepository,
+            StatusUpdateService statusUpdateService,
             IDbContextResolver contextResolver
-            ) : base(loggerFactory, jsonApiContext, currentUserRepository, contextResolver)
+            ) : base(loggerFactory, jsonApiContext, currentUserRepository, statusUpdateService, contextResolver)
         {
         }
-        public override IQueryable<Organization> Filter(IQueryable<Organization> query, FilterQuery filterQuery)
+        public override IQueryable<Organization> Filter(IQueryable<Organization> entities, FilterQuery filterQuery)
```
NullVoxPopuli: why all these types of changes?

source/OptimaJet.DWKit.StarterApplication/Repositories/BaseRepository.cs
chrisvire: Yeah, I wasn't sure what to do in this case where DeleteAsync returned false.  Did it really not get deleted?

source/OptimaJet.DWKit.StarterApplication/Repositories/OrganizationRepository.cs
chrisvire: There were warnings in visual studio that the name didn't match the original.

source/OptimaJet.DWKit.StarterApplication/Repositories/BaseRepository.cs
NullVoxPopuli: I feel like there should be an exception or something if something goes wrong?
would probably need to comb through the source to see what's possible

source/OptimaJet.DWKit.StarterApplication/Models/IStatusUpdate.cs
```diff
@@ -0,0 +1,9 @@
+﻿using System.Collections.Generic;
+
+namespace OptimaJet.DWKit.StarterApplication.Models
+{
+    public interface IStatusUpdate
```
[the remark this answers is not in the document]
chrisvire: Originally, `IStatusUpdate` just had `UpdateGroups` to know which groups should be notified.  Then I had to add StringId since some Model objects might inherit from either `Identifiable` or `Identifiable<Guid>` and `StringId` was common to both interfaces.  So I added it to `IStatusUpdate`.  At another time, it had `EntryEntity` so that we could get the `TableName` when the implementation was in `AppDbContext`.  Since we moved it to the Repository implementation, we hav to lookup the TableName from the contained `DbContext`.

So IStatusUpdate is not a create name.  We need some way to mark a model object so that we know that we want to send out notifications (because it isn't free).

source/OptimaJet.DWKit.StarterApplication/Repositories/Background/JobRepository.cs
chrisvire: @NullVoxPopuli: So if I change it to `EntityHookService` and `IHookEntity`?

source/OptimaJet.DWKit.StarterApplication/Repositories/BaseRepository.cs
chrisvire: I looked through the source and slacked you why it was like that (at least in JsonApiDotNetCore).  bool is just whether the entity id was found.  The Controller returns NotFound if false and NoContent if true.

source/OptimaJet.DWKit.StarterApplication/Models/IStatusUpdate.cs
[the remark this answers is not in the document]
chrisvire: I found I could use `IIdentifiable` so I don't need `IStatusUpdate` or `IEntityHook` (or any additional interface to mark which to check).  Now everything is done through DI.

PR #515, opened 2019-02-04
CI Updates

NullVoxPopuli:
The original intent behind the original set of updates was to:
 - make it so we could run the scripts locally
 - have the scripts be idempotent, so that if you already have the deps installed, it doesn't try to re-install them
 - reuse a bunch of variables that may have been previously duplicated
 - provide more messaging in the logs to see what is going on
 - and most importantly, ensure that `docker build` isn't going to fail due to dep changes within either project (frontend and api)

However, I made the mistake of not scrubbing tag names for illegal characters when tagging docker images. 
Tagging and pushing doesn't actually need to happen for the non-deployable branches, so in this PR, I've added a condition to only push and tag on master and develop.

### DIFF
--- a/Dockerfile.nginx
+++ b/Dockerfile.nginx
@@ -1,10 +1,15 @@
 # ---- Base ----
 FROM nginx:1.15.8-alpine AS base
+# nginx 1.15.8 uses alpine 3.9
+#
+# so here is the nodejs search for that version of alpine 
+# https://pkgs.alpinelinux.org/packages?name=nodejs&branch=v3.9&repo=main
+#
 RUN apk add --no-cache git curl
 RUN apk add --no-cache \
       --repository http://dl-3.alpinelinux.org/alpine/edge/community \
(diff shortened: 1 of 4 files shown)


Dockerfile.nginx
```diff
@@ -1,10 +1,15 @@
...
+# so here is the nodejs search for that version of alpine 
+# https://pkgs.alpinelinux.org/packages?name=nodejs&branch=v3.9&repo=main
+#
 RUN apk add --no-cache git curl
 RUN apk add --no-cache \
       --repository http://dl-3.alpinelinux.org/alpine/edge/community \
-      nodejs=8.14.0-r0 \
-      npm=8.14.0-r0 \
+      nodejs=10.14.2-r0 \
+      npm=10.14.2-r0 \
```
NullVoxPopuli: node 8 is no longer supported on alpine 3.9, needed to upgrade

PR #529, opened 2019-02-12
33105 - Group dropdown didn't fully close after click

NullVoxPopuli:
both the Role and Group dropdowns on the user table didn't fully close after changing the options if the dropdown happened to extend belong the bottom of the viewport.

Turns out, the SemanticUI dropdown was just configured wrong.


source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/project-directory/list/__tests__/column-selector-acceptance-test.tsx
```diff
@@ -45,6 +45,7 @@ describe('Acceptance | Project Directory | Column selector', () => {
   describe('navigates to project directory page', () => {
     beforeEach(async function() {
       await visit('/directory');
+      await when(() => assert(page.rows().length > 0, `Expected project table to be rendered`));
```
NullVoxPopuli: sanity checks everywhere.
async testing is full of doubts. o.o

PR #533, opened 2019-02-13
31008 display product status in project details view

NullVoxPopuli:
- fix error rendering

image

the activity name is currently being used, which doesn't make a whole lot of sense for this sentence.
idk if we want to come up with a translation map based on the activityName or Status. or... ?? idk


source/SIL.AppBuilder.Portal.Frontend/package.json
```diff
@@ -71,8 +71,8 @@
     "react-notify-toast": "^0.5.0",
     "react-orbitjs": "github:DeveloperTown/react-orbitjs#9913bdeac3495a9f75413c30f168ed617cf29124",
     "react-redux": "^6.0.0",
-    "react-router": "^4.3.1",
-    "react-router-dom": "^4.3.1",
+    "react-router": "^4.4.0-beta.6",
```
NullVoxPopuli: in order to get access to the router context via private api to make a `useRouter` hook.

source/SIL.AppBuilder.Portal.Frontend/src/lib/auth/get-current-permissions.ts
```diff
@@ -5,15 +5,15 @@ import Store from '@orbit/store';
...
+import { useCurrentUser } from '@data/containers/with-current-user';
 
 interface IHasRoleInOrganizationArgs {
   organization?: OrganizationResource;
   roleName: ROLE;
 }
 
 export function getPermissions(organization?: OrganizationResource) {
-  const { currentUser } = useContext(CurrentUserContext);
+  const { currentUser } = useCurrentUser();
```
dalinarkholin: awesome. consider file name change?

NullVoxPopuli: ah yeah, it should be `get-permissions`

PR #536, opened 2019-02-15
Use Button.href instead of embedding Link

chrisvire:


chrisvire:
@NullVoxPopuli: I am not sure how I would add a test to make sure that the link of the `<IconButton>` was an external link (if that would be a good test).

NullVoxPopuli:
The only way to test it that way is to check if the `href` attribute begins with `https://{whatever-domain}`

PR #544, opened 2019-02-19
Claim ownership: *the sequel*

NullVoxPopuli:


source/OptimaJet.DWKit.StarterApplication/Services/Workflow/WorkflowProjectService.cs
```diff
@@ -0,0 +1,60 @@
...
+          var previousOwner = await this.UserRepository.GetAsync(previousOwnerId);
+          var newOwner = await this.UserRepository.GetAsync(newOwnerId);
+
+          project.Products.ForEach(product => {
+              this.ReassignUserTasksForProduct(product, previousOwner, newOwner);
+          });
+        }
+
+        private void ReassignUserTasksForProduct(Product product, User previous, User next) {
+          product.UserTasks
```
NullVoxPopuli: @chrisvire what else has to change?

I was looking at the dashboard at http://localhost:7081/
and the task in there appropriately updates

Also, I suppose this type of reassignment is a bit naiive, as it assumes both users have the same permissions.
It's totally possible that the previous user is an org admin, and would retain some tasks, just because they are an org admin, yeah?

NullVoxPopuli: Would it be better to, using code from the WorkflowProductService, to remove and then re-create? rather than re-assign?

```cs
// parts of 
ProductProcessChangedAsync(...)
// may be relevant
```

NullVoxPopuli: yeah actually, it looks like if I can create a `ProductProcessChangedArgs`, I should just be able to reuse that method entirely

NullVoxPopuli:
I'm having some D.I. config issues atm, but the resulting logic should more or less be the same

source/OptimaJet.DWKit.StarterApplication/Models/Product.cs
```diff
@@ -21,6 +21,10 @@ public class Product : Identifiable<Guid>, ITrackDate
         [HasMany("user-tasks", Link.None)]
         public virtual List<UserTask> UserTasks { get; set; }
```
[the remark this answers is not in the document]
NullVoxPopuli: I'm unsure of what makes the most sense w/r/t those two.
presently, the frontend maps from product.

source/OptimaJet.DWKit.StarterApplication/Models/Product.cs
```diff
@@ -21,6 +21,10 @@ public class Product : Identifiable<Guid>, ITrackDate
...
 
+        public virtual WorkflowProcessInstance WorkflowProcessInstance { get; set; }
```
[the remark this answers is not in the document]
NullVoxPopuli: I'll take a look!

source/OptimaJet.DWKit.StarterApplication/Services/Workflow/WorkflowProductService.cs
```diff
@@ -184,8 +187,37 @@ public async Task ProductProcessChangedAsync(ProductProcessChangedArgs args)
             }
         }
 
+        public async Task ReassignUserTasksForProduct(Product product) {
```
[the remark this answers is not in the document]
NullVoxPopuli: >  In ReassignUserTasksForProduct, we would want to capture the comment from the previous UserTask(s).

I had originally done that (this commit: https://github.com/sillsdev/appbuilder-portal/pull/544/commits/59d8355cc50d22f0475aa2b7a85b96a7acb585aa#diff-d2deae5122d392cecaf0f41e802ea812R70), but ran in to a situation where the owner change may have different permissions?, so then it wouldn't make sense to just 1-1 copy over the previous user tasks. However, if that's not the case, maybe, by convention of role, we _can_ say that we can just copy over (or just change the user Id).

source/OptimaJet.DWKit.StarterApplication/Models/Product.cs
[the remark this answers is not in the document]
chrisvire: Nevermind.  I think it is best to leave it.

source/OptimaJet.DWKit.StarterApplication/Services/Workflow/WorkflowProductService.cs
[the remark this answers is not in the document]
chrisvire: The code isn't aware why the UserTasks are assigned (owner, orgadmin or others in the future).  These are determined by the available commands in the workflow.  So it isn't safe to do reassignment of tasks.  It would be good to keep the comment from the previous command execution.

Now upon reflection, the comments really should be on the completed `ProductionTransition`.  That can be fixed at a later time.  I have created https://developertown.visualstudio.com/SIL%20International%20Scriptoria/_workitems/edit/33431 to track this part.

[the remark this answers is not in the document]
chrisvire: If it would make it cleaner to implement to save the comment to ProductionTransition, then feel free to make that change.

source/OptimaJet.DWKit.StarterApplication/Models/Product.cs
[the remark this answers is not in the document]
NullVoxPopuli: ProductWorkflow doesn't exist in the project
```bash
$ git grep "ProductWorkflow" | wc -l
0
```
how do you access ProductWorkflow  / the data from `WorkflowProcessInstance`? :-\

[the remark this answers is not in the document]
chrisvire: @NullVoxPopuli : Yeah, I know it doesn't exist.  I was hoping that would be class name used in Scriptoria.  If it has to be WorkflowProcessInstance to do the data mapping with EF, I guess that is OK.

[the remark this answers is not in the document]
NullVoxPopuli: oh, you were just wanting it to be renamed.

e
z
:+1:

source/OptimaJet.DWKit.StarterApplication/Services/Workflow/WorkflowProductService.cs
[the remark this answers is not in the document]
chrisvire: It seems like there could be more shared code with just the difference in how Comments are handled.  I didn't see pulling the Comment from the previous UserTasks.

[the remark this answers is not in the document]
NullVoxPopuli: I don't know if there is a way to pull the comment from each user task, because there is a chance for non-equal task conversion. We should start with 4 tasks and end up with 3 or 5. Is the comment the same for all tasks? (that's the only way I see carrying the comment over working :-\ )

source/OptimaJet.DWKit.StarterApplication/Services/Workflow/WorkflowProductService.cs
```diff
@@ -141,21 +137,36 @@ public async Task ProductProcessChangedAsync(ProductProcessChangedArgs args)
 
             await RemoveTasksByProductId(product.Id);
 
+            var tasks = await ReassignUserTasksForProduct(product);
```
chrisvire: `ReassignUserTasksForProduct` calls `GetCurrentTaskCommnt` and then `RemoveTasksByProductId`.  It looks lik the call to `RemoveTasksByProductId` above this should be removed.

NullVoxPopuli: oh good catch!

source/OptimaJet.DWKit.StarterApplication/Services/Workflow/WorkflowProductService.cs
```diff
@@ -133,29 +128,43 @@ public async Task ProductProcessChangedAsync(ProductProcessChangedArgs args)
...
+            // Clear the WorkflowComment
+            if (!String.IsNullOrWhiteSpace(product.WorkflowComment))
+            {
+                // Clear the comment
+                product.WorkflowComment = "";
+                await ProductRepository.UpdateAsync(product);
+            }
+        }
+
+        private async Task SendNotifications(List<UserTask> tasks, Product product, ProductProcessChangedArgs args)
```
chrisvire: I am really confused by the `SendNotifications` function.  Why is is running the `Runtime.GetAllActorsForDirectCommandTransitions` again?  

It should be sending out a Notification for each input task (which has a user associated with it).
It doesn't need the args.

source/OptimaJet.DWKit.StarterApplication/Services/Workflow/WorkflowProductService.cs
```diff
@@ -133,29 +128,43 @@ public async Task ProductProcessChangedAsync(ProductProcessChangedArgs args)
...
             if (product == null)
             {
                 Log.Error($"Could find Product for ProcessId={args.ProcessId}");
                 return;
             }
 
-            await RemoveTasksByProductId(product.Id);
+            var tasks = await ReassignUserTasksForProduct(product);
+
+            await SendNotifications(tasks, product, args);
```
chrisvire: The `SendNotifications` should be happening inside of `ReassignUserTasksForProduct` since notification also needs to happen for your new situation of reassigning the project to a different user.

run
```diff
@@ -119,6 +121,12 @@ function runstuff {
         && echo "DB setup with sample data"
     ;;
 
+    bootstrap:ui-static)
```
chrisvire: `source/SIL.AppBuilder.Portal.Frontend/src/public/assets/language/ ` needs to be added to `.gitignore`

NullVoxPopuli: it is in the frontend gitignore

source/OptimaJet.DWKit.StarterApplication/Services/Workflow/WorkflowProductService.cs
NullVoxPopuli: ok, I was actually wondering about that and forgot to ask if we should spam notifications when this happens :p

source/OptimaJet.DWKit.StarterApplication/Services/Workflow/WorkflowProductService.cs
NullVoxPopuli: Cause I've been a hot mess the past couple days. Fix incoming!

source/OptimaJet.DWKit.StarterApplication/Services/Workflow/WorkflowProductService.cs
```diff
@@ -133,59 +124,126 @@ public async Task ProductProcessChangedAsync(ProductProcessChangedArgs args)
...
+
+                await SendNotificationForTask(userTask, user, product, args);
+
+                result.Add(createdUserTask);
+            }
+
+            return result;
+        }
+
+        private async Task SendNotificationForTask (UserTask task, User user, Product product, ProductProcessChangedArgs args = null)
```
NullVoxPopuli: passings `args` through was initially/partially the reason I didn't more cleanly refactor things at first.

source/OptimaJet.DWKit.StarterApplication/Services/Workflow/WorkflowProjectService.cs
```diff
@@ -0,0 +1,49 @@
...
+            WorkflowProductService productService
+        )
+        {
+            ProjectRepository = projectRepository;
+            WorkflowProductService = productService;
+        }
+
+        public async Task ReassignUserTasks(int projectId) {
+            var project = this.ProjectRepository.Get()
+                .Include(proj => proj.Products)
```
NullVoxPopuli: I wish EF had a more ergonomic way to do this. :-\

chrisvire: Are you talking about the `.Include` syntax?  Yeah, I agree.

NullVoxPopuli: yeah, and having to do .Include multiple times if you want to go down sibling branches. I think I could do a static extension to wrap something that ends up doing what I've done here. but :man_shrugging:  maybe next time :D

source/OptimaJet.DWKit.StarterApplication/Services/Workflow/WorkflowProductService.cs
chrisvire: I don't know why `args` is a parameter.
* you can remove it from the Log statement (or move the Log statement to `ProductProcessChangedAsync` where it is more relevant).
* fromActivity is not used in the notification so that entry in the messageParms can be dropped.

source/OptimaJet.DWKit.StarterApplication/Services/Workflow/WorkflowProductService.cs
```diff
@@ -107,24 +107,15 @@ protected async Task CreateWorkflowProcessInstance(Product product)
...
-            var emptyItems = ProductTransitionRepository.Get()
-                .Where(pt => pt.WorkflowUserId == null && pt.ProductId == args.ProcessId)
-                .Select(pt => pt.Id).ToList();
-            foreach (var item in emptyItems)
-            {
-                await ProductTransitionRepository.DeleteAsync(item);
-            }
-            // Create PreExecute entries
-            await Runtime.PreExecuteFromCurrentActivityAsync(args.ProcessId);
+            await RecreatePreExecuteEntries(args.ProcessId);
```
chrisvire: `ClearPreExecuteEntries should be at the beginning of `ReassignUserTasksForProduct`.  A `CreatePreExecuteEntries` should be at the end of `ReassignUserTasksForProduct`.

source/OptimaJet.DWKit.StarterApplication/Services/Workflow/WorkflowProductService.cs
NullVoxPopuli: args is a param because it contained information that isn't used in the tasks or notifications
 - ExecutingCommand
 - Used on creation, but not re-assignment (due to there just not being a ProcessChange (from DWKit)
   - PreviousActivityName
   - CurrentState

fromActivity _IS_ used in the notification, in the middle of the Dictionary

> moving log to ProductProcessChangedAsync
 - we'd get one log for the whole thing, instead of per-task, but at this point it feels like bike-shedding, / also I think I'd prefer to just remove the log, and pass PreviousActivityName for the notification explicitly.

source/OptimaJet.DWKit.StarterApplication/Services/Workflow/WorkflowProductService.cs
NullVoxPopuli: done

PR #555, opened 2019-02-28
fix a bunch of stuff

NullVoxPopuli:

PR #557, opened 2019-03-01
Add locale for reviewer.

davidmoore1:
Changed existing locale-select to locale-switch because it better described the existing functionality and locale-select seemed to be the appropriate name for the new dropdown.
Created new locale-select to select an entry from the list of available UI languages and then return
it to the caller.


NullVoxPopuli:
blocked by: https://github.com/sillsdev/appbuilder-portal/pull/555

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/inputs/locale-select/index.tsx
```diff
@@ -22,39 +22,54 @@ export type IProps = IOwnProps & WithDataProps & i18nProps;
...
-    const { currentUser, i18n } = this.props;
+    const { currentUser, i18n, onChange } = this.props;
     const attributes = attributesFor(currentUser) as UserAttributes;
     const userLocale = attributes.locale;
     const { options, language } = i18n;
     const languages = Object.keys(options.resources);
 
-    const selected = userLocale || language;
-
+    this.defaultValue = userLocale || language;
```
NullVoxPopuli: in react components, we aren't supposed to use class properties to manage state (React is the only JS library that I know of to require this)

you'll need to use state / setState.

Some info here: https://reactjs.org/docs/state-and-lifecycle.html#using-state-correctly

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/inputs/locale-switch/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+        onChange={this.onSelect}
+      />
+    );
+  }
+}
+
+export default compose(
+  withTranslations,
+  withCurrentUserContext
+)(LocaleSwitch);
```
NullVoxPopuli: for new components going forward (as of two weeks ago anyway), 
to get access to translations and current user, we should use "hooks":
like this:

```
export default function LocaleSwitch({ onChange, className } IProps) {
  const { currentUser } = useCurrentUser();
  const { i18n } = useTranslations();

  const onSelect = ...

  // the rest of the variables

  return (
    // same as in your render function, but no `this`
  );
}
```

this is because of a shift in the react programming paradigm, where classes are no longer used for components and higher-order components (withTranslations, etc) are used MUCH less frequently

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/inputs/locale-select/index.tsx
```diff
@@ -22,39 +22,54 @@ export type IProps = IOwnProps & WithDataProps & i18nProps;
...
+    this.defaultToFirst();
+  }
+
+  defaultToFirst = () => {
+    const { onChange, selected } = this.props;
+    if (!selected) {
+      onChange(this.defaultValue);
+    }
+  };
+
```
NullVoxPopuli: as a function-component, this could be converted like:
```ts
export default function LocaleSelect({ selected, onChange}: IProps) {
  const onSelect = async (e, { value }) => {
    e.preventDefault();

     if (value === selected) {
       return;
     }

     onChange(value);
  }

  // ...

  return (
    <Dropdown defaultValue={selected} ...otherstuff />
   );
}
```

Though, looking at those closer, I must ask: what is the point of the default value property?

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/inputs/locale-select/index.tsx
davidmoore1: Redid this significantly, including the state change requests mentioned here

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/inputs/locale-switch/index.tsx
davidmoore1: Rewrote this to match the comment

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/inputs/locale-select/index.tsx
davidmoore1: I found that I could significantly simplify this class.  It now has a first time variable so that it can set the value of the variable in the caller for the case where no selection is made initially, which is kind of what it was trying to do before with the default value.

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/inputs/locale-select/index.tsx
```diff
@@ -17,44 +17,57 @@ export interface IOwnProps {
...
     const attributes = attributesFor(currentUser) as UserAttributes;
     const userLocale = attributes.locale;
     const { options, language } = i18n;
     const languages = Object.keys(options.resources);
-
-    const selected = userLocale || language;
-
+    const defaultLocale = userLocale || language;
+    if (firstTime) {
+      this.setState({ firstTime: false });
```
NullVoxPopuli: maybe you want this? https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops

but also: https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html

NullVoxPopuli: I guess I don't understand why there is special firstTime logic :-\

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/inputs/locale-switch/index.tsx
```diff
@@ -0,0 +1,61 @@
...
+      attributes: { locale: value },
+    });
+  };
+
+  const attributes = attributesFor(currentUser) as UserAttributes;
+  const userLocale = attributes.locale;
+  const { options, language } = i18n;
+  const languages = Object.keys(options.resources);
+
+  const selected = userLocale || language;
```
NullVoxPopuli: :+1:

source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/projects/show/overview/reviewers/form.tsx
```diff
@@ -50,6 +57,19 @@ class AddReviewerForm extends React.Component<IProps> {
     });
   };
 
+  setEmptyLocale = () => {
```
NullVoxPopuli: this method doesn't do any setting.

source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/projects/show/overview/reviewers/form.tsx
```diff
@@ -50,6 +57,19 @@ class AddReviewerForm extends React.Component<IProps> {
...
+    const { currentUser, i18n } = this.props;
+    const { locale } = this.state;
+    var defaultLocale = locale;
```
NullVoxPopuli: let should be used here instead of `var`.
var has scoping issues https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/let

source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/projects/show/overview/reviewers/form.tsx
```diff
@@ -50,6 +57,19 @@ class AddReviewerForm extends React.Component<IProps> {
     });
   };
 
+  setEmptyLocale = () => {
+    const { currentUser, i18n } = this.props;
+    const { locale } = this.state;
```
NullVoxPopuli: this could be simplified to:

`let { locale: defaultLocale } = this.state`

source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/projects/show/overview/reviewers/form.tsx
```diff
@@ -50,6 +57,19 @@ class AddReviewerForm extends React.Component<IProps> {
...
+    var defaultLocale = locale;
+    if (isEmpty(locale)) {
+      const attributes = attributesFor(currentUser) as UserAttributes;
+      const userLocale = attributes.locale;
```
NullVoxPopuli: this could be simplified to:
`const { locale: userLocale } = attributesFor(...`

source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/projects/show/overview/reviewers/form.tsx
```diff
@@ -50,6 +57,19 @@ class AddReviewerForm extends React.Component<IProps> {
...
+      const { language } = i18n;
```
NullVoxPopuli: since language is only used once here, and since the property-access path isn't very long, you probably don't need to destructure it, and could just do

`defaultLocale = userLocale || i18n.language`

PR #567, opened 2019-03-08
Trying out some bugfixes

NullVoxPopuli:
Fixes:
 - issue where removing a product would cause the UI to crash
 - adds yarn to the toolchain
 - a bug where a non-deleted resource did not exist at the time of trying to send an update from the api to the ui (due to a race condition due to the async-nature of the background jobs)
 - a bunch of 404 errors in the console due to the i18next library trying to retrieve everything from the backend, even though the resources were already loaded at build time
 - finally be able to replicate a bug that looks exactly like what @liztabor found the other day, and fix that.


source/SIL.AppBuilder.Portal.Frontend/src/sockets/data-socket.tsx
```diff
@@ -21,23 +21,50 @@ import { DataClient } from './clients';
 
 import { dataToLocalCache } from '~/data/orbitjs-operations-support/serialize-from-api';
 
+import { useMemoIf } from '~/lib/hooks';
+
+import { DataHub } from './clients/data';
+
+const mockClient: DataClient = {
```
NullVoxPopuli: ugly hack for tests, :(

NullVoxPopuli: The library we're using for sockets has no tests as well: https://github.com/sketch7/signalr-client/blob/master/test/test-setup.ts
woo

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/header/notifications/display.tsx
```diff
@@ -1,91 +1,92 @@
...
+
+  const haveAllNotificationsBeenSeen = notifications.every(
+    (n) => attributesFor(n).dateRead !== null
+  );
+  const hasNotifications = notifications.length > 0;
+
+  const toggle = () => {
+    if (visible) {
+      markAllAsViewed();
+      refetch();
```
NullVoxPopuli: this comes for the `query` HoC

source/SIL.AppBuilder.Portal.Frontend/tests/mocha.opts
```diff
@@ -1,3 +1,3 @@
 {src,tests}/**/*-test.{ts,tsx}
---timeout 3000
+--timeout 5000
```
NullVoxPopuli: travis is slow

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/header/notifications/display.tsx
```diff
@@ -1,91 +1,92 @@
...
+      <div className={`ui menu transition notifications ${isMenuVisible}`}>
+        {hasNotifications && (
+          <>
+            <div className='notification-buttons'>
+              <a href='#' data-test-clear-all onClick={preventDefault(clearAll)}>
+                {t('header.clearAll')}
+              </a>
+            </div>
+
+            <div className={notifications.length > 3 ? 'scrollable-menu' : ''}>
```
abillingsley: what is scrollable-menu doing? Why wouldn't we just apply it always and let the browser overflow if necessary?

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/header/notifications/display.tsx
```diff
@@ -1,91 +1,92 @@
...
-          {!hasNotifications && (
-            <div className='notification-no-data'>{t('header.emptyNotifications')}</div>
-          )}
-        </div>
+  const isMenuVisible = visible ? 'visible' : '';
+
+  return (
+    <div className='ui top right pointing dropdown' data-test-header-notification>
+      <div
+        style={{ zIndex: 1 }}
```
abillingsley: should this be a part of the full-overlay styling?

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/header/notifications/display.tsx
```diff
@@ -1,91 +1,92 @@
...
-                ))}
-              </div>
-            </>
-          )}
-
-          {!hasNotifications && (
-            <div className='notification-no-data'>{t('header.emptyNotifications')}</div>
-          )}
-        </div>
+  const isMenuVisible = visible ? 'visible' : '';
```
abillingsley: a variable name like `isMenuVisible` makes me think boolean but instead, this is defining the class to apply

source/SIL.AppBuilder.Portal.Frontend/src/translations/index.tsx
```diff
@@ -14,7 +14,10 @@ i18n
   // https://github.com/i18next/i18next-browser-languageDetector
   .use(LanguageDetector)
   .use(initReactI18next)
-  .use(XHR)
+  // .use(XHR)
```
abillingsley: I get from the comment that we want to do local and remote translations.  Double checking that it is ok to turn off the remote translations until the plugin is implemented?

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/header/notifications/display.tsx
NullVoxPopuli: totally agree. I didn't implement this logic, I just refactored the component from a class to function

source/SIL.AppBuilder.Portal.Frontend/src/translations/index.tsx
NullVoxPopuli: at the moment we aren't using any remote translations, but it's planned for 'eventually'

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/header/notifications/display.tsx
NullVoxPopuli: it should at least have a default. I didn't write this though. the original code is from here: https://github.com/sillsdev/appbuilder-portal/pull/567/files#diff-26e4c95cb718005691cad1c85e16de6bL51

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/header/notifications/display.tsx
NullVoxPopuli: totally agree.
It was just pulled from here: https://github.com/sillsdev/appbuilder-portal/pull/567/files#diff-26e4c95cb718005691cad1c85e16de6bL46

PR #578, opened 2019-03-18
Convert from alltags.json to langtags.json

chrisvire:
* WritingSystemsTechnology is switching the alltags.json to
  langtags.json
* regions attribute is changing from being a space separated list
  in a string to being an array
* New file is https://raw.githubusercontent.com/silnrsi/sldr/master/extras/langtags.json
* Include known variants (not sure how this should be used in
  suggestions)

Issue:
* in /projects/new, I get error "Cannot read property
'toLocaleLowerCase' of undefined
** this is in src/ui/components/input/locale-input/-utils/helpers.tsx,
   line 136
* VCode thinks there is an error in .../locale-input/-utils/localize.ts
** I remove info.regions.split(' ') since regions is a string[] instead
   of string.  I thought the existing .map() would work in string[]

### DIFF
--- a/source/SIL.AppBuilder.Portal.Frontend/scripts/split-langtags-json.js
+++ b/source/SIL.AppBuilder.Portal.Frontend/scripts/split-langtags-json.js
@@ -1,0 +1,23 @@
+const fs = require('fs');
+
+const [_executor, _file, inputName, output1Name, output2Name] = process.argv;
+
+let raw = fs.readFileSync(inputName);
+let data = JSON.parse(raw);
+
+let output1 = data.filter(function (el){
+  return !el.tag.startsWith("_")
+});
+let output1String = JSON.stringify(output1,null,2);
+
+let output2 = data.filter(function (el){
+  return el.tag.startsWith("_")
+});
+
+let output2String = JSON.stringify(output2, null, 2);
+
+fs.writeFileSync(output1Name, output1String);
+
+fs.writeFileSync(output2Name, output2String);
+
+console.log(`JSON output at ${output1Name} and ${output2Name}`);
--- a/source/SIL.AppBuilder.Portal.Frontend/src/ui/components/inputs/locale-input/-utils/helpers.tsx
+++ b/source/SIL.AppBuilder.Portal.Frontend/src/ui/components/inputs/locale-input/-utils/helpers.tsx
@@ -16,13 +16,15 @@ export const getSuggestions = (data: ILanguageInfo[]) => (value) => {
   const has = hasSearchTerm(value);
 
   return data.filter(
-    ({ name, nameInLocale, region, tag, localname, regions, names }) =>
+    ({ name, nameInLocale, region, tag, localname, regions, names, variants }) =>
       has(nameInLocale) ||
       has((names || []).join()) ||
       has(localname || name) ||
-      has(regions || region) ||
+      has((regions || []).join()) ||
+      has(region) ||
+      has((variants || []).join()) ||
       has(tag)
-  );
+   );
 };
 
 export const getSuggestionValue = (suggestion: ILanguageInfo) => suggestion.tag;
--- a/source/SIL.AppBuilder.Portal.Frontend/src/ui/components/inputs/locale-input/-utils/helpers.tsx
+++ b/source/SIL.AppBuilder.Portal.Frontend/src/ui/components/inputs/locale-input/-utils/helpers.tsx
@@ -24,7 +24,7 @@ export const getSuggestions = (data: ILanguageInfo[]) => (value) => {
       has(region) ||
       has((variants || []).join()) ||
       has(tag)
-   );
+  );
 };
 
 export const getSuggestionValue = (suggestion: ILanguageInfo) => suggestion.tag;
--- a/source/SIL.AppBuilder.Portal.Frontend/src/ui/components/inputs/locale-input/-utils/localize.ts
+++ b/source/SIL.AppBuilder.Portal.Frontend/src/ui/components/inputs/locale-input/-utils/localize.ts
@@ -36,7 +36,6 @@ export function localizeTagData(data: ILanguageInfo[], t): ILanguageInfo[] {
         info.regions &&
         info.regions
           .map((region) => tryLocalize('territories', region, region))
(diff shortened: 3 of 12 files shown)


abillingsley:
@chrisvire the first problem I am noticing is that the langtags.json now includes two entries like

```
    {
        "tag": "_globalvar", 
        "variants": [ "simple" ]
    }, 
    {
        "tag": "_phonvar", 
        "variants": [ "alalc97", "fonipa", "fonkirsh", "fonnapa", "fonupa", "fonxsamp" ]
    }, 
```

a fully formed entry looks something like
```
    {
        "full": "aa-Latn-ET", 
        "iana": "Afar", 
        "iso639_3": "aar", 
        "localname": "Qafar", 
        "name": "Afar", 
        "names": [ "Afaraf", "Qafar", "Qafar af", "ʿAfár af" ], 
        "region": "ET", 
        "regionname": "Ethiopia", 
        "regions": [ "DJ", "ER" ], 
        "sldr": true, 
        "tag": "aa", 
        "tags": [ "aa-ET", "aa-Latn" ]
    }, 
```

as a result a number of the fields are undefined which is what is causing the 
> "Cannot read property 'toLocaleLowerCase' of undefined"


basically we have defined an api contract requiring some number of fields and those two entries are in violation of that contract.  I am not sure how much control we have over this file.  Depending on the answer a couple options come to mind

1. preprocess this the langtag.json files as part of refresh_languages.sh to include only valid entries
2. leave refresh_languages.sh unchanged and perform a filter client side after fetching the file so the array of entiries only contains valid entries

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/inputs/locale-input/-utils/localize.ts
```diff
@@ -27,14 +27,14 @@ export function localizeTagData(data: ILanguageInfo[], t): ILanguageInfo[] {
...
       nameInLocale: tryLocalize('languages', info.tag, info.name),
 
       // These are the only fields that need localization
       region: info.region && tryLocalize('territories', info.region, info.region),
       regions:
         info.regions &&
         info.regions
-          .split(' ')
           .map((region) => tryLocalize('territories', region, region))
           .join(', '),
```
abillingsley: the type definition for ILanguageInfo previously defined regions as a string and now defines regions an an array of string
```
-  regions?: string;
+ regions?: string[];
```

```
.join(', ')
``` 
turns an array of string regions into a comma delimited string of regions which is not how the type is defined. which causes some issues in `helpers.tsx` since an array of string is expected based on the type but at runtime the property is made a string

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/inputs/locale-input/-utils/helpers.tsx
```diff
@@ -16,13 +16,15 @@ export const getSuggestions = (data: ILanguageInfo[]) => (value) => {
   const has = hasSearchTerm(value);
 
   return data.filter(
-    ({ name, nameInLocale, region, tag, localname, regions, names }) =>
+    ({ name, nameInLocale, region, tag, localname, regions, names, variants }) =>
       has(nameInLocale) ||
       has((names || []).join()) ||
       has(localname || name) ||
-      has(regions || region) ||
+      has((regions || []).join()) ||
```
abillingsley: this is not really an array at runtime see comment below

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/inputs/locale-input/-utils/localize.ts
abillingsley: if you remove the `.join` a new error is thrown in `find-additional-matches.tsx`

I _think_ you can replace this code 

```
    let value = Array.isArray(rawValue) ? rawValue.join(', ') : rawValue;
    let num = Array.isArray(rawValue) ? rawValue.length : 1;

    if (rawValue && key === 'regions') {
      num = rawValue.split(' ').length;
    }
```

with this code
```
    const value = Array.isArray(rawValue) ? rawValue.join(', ') : rawValue;
    const num = Array.isArray(rawValue) ? rawValue.length : 1;
```

since regions is now an array we no longer need special handling to compute the count.  I am not sure if this will cause regressions anywhere else but this does seem to fix the problem for me

abillingsley:
@chrisvire this PR https://github.com/sillsdev/appbuilder-portal/pull/579 assumes a valid langtag.json file (manually modified mine locally to remove bad entries)

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/inputs/locale-input/-utils/localize.ts
chrisvire: Yes, we figured out that we needed to remove the `.join`.

Yes, we also remove the special case for regions in `find-additional-matches.tsx.

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/inputs/locale-input/-utils/helpers.tsx
chrisvire: I believe it is an array now.

chrisvire:
> @chrisvire this PR #579 assumes a valid langtag.json file (manually modified mine locally to remove bad entries)

Yep.  There is a new format for the JSON that includes those entries at the top that apply to all languages.  Sigh.

source/SIL.AppBuilder.Portal.Frontend/scripts/split-langtags-json.js
```diff
@@ -0,0 +1,23 @@
...
+let output1 = data.filter(function (el){
+  return !el.tag.startsWith("_")
+});
+let output1String = JSON.stringify(output1,null,2);
+
+let output2 = data.filter(function (el){
+  return el.tag.startsWith("_")
+});
+
+let output2String = JSON.stringify(output2, null, 2);
```
abillingsley: instead of defining all these variables as `let` use `const` since the variable is never being reassigned

PR #588, opened 2019-03-25
use additional condition to check for failures

NullVoxPopuli:
When the test script was created, the results of stdout were piped to a file -- this made the exit code always return 0, as the pipe always succeeded.

This PR adds a check to that piped file to see if any tests failed.


davidmoore1:
Kind of an unusual test sequence, where the test failing indicates that the code change is good.  :-)

NullVoxPopuli:
now to fix the test :)

NullVoxPopuli:
actually, I think the fix for this failing test is in my other branch
_hmm_

NullVoxPopuli:
pending passing tests on: https://github.com/sillsdev/appbuilder-portal/pull/568

PR #593, opened 2019-03-29
Workflow / Workflow-Socket Bugfixes / Hacks

NullVoxPopuli:
The DWKit websocket Hub is hard coded at `hubs/notifications`, even though the hub is used for way more than notifications.  When notifications were implemented, a socket hub called `hubs/notifications` was created -- which makes sense, cause... notifications _would_ be sent over this socket. Since we now have `{ json:api }` over websocket now, and with the background hook notifier, we no longer need the `ScriptoriaHub` and we can delete all that code to just rely on the data / jsonapi hub, freeing up the hub namespace DWKit hardcoded to be used for DWKit things.

Also
 - fixes an issue with notifications where if an exception occurs when deleting/clearing a notification, the rest of the notifications cannot be cleared.
 - ~~fixes: https://developertown.visualstudio.com/SIL%20International%20Scriptoria/_workitems/edit/34286 (version sort order)~~ Moved to https://github.com/sillsdev/appbuilder-portal/pull/596
 - fixes an issue with websocket connections not being tied to users (most of the C# code) and resulted in some something showing up in multiples (such as notifications) https://developertown.visualstudio.com/SIL%20International%20Scriptoria/_workitems/edit/34490
 - ~~quick UI/UX fix for: https://developertown.visualstudio.com/SIL%20International%20Scriptoria/_workitems/edit/34492~~ moved to https://github.com/sillsdev/appbuilder-portal/pull/597
 - Adds live data updating to the tasks view
   - TODO: need to add the ability for subscriptions to pull related records when they update (`include`)


NullVoxPopuli:
Merged without external review due to how important some of the UX fixes are, and time constraints :(

PR #596, opened 2019-04-03
When a build artifact version is null, show that the build is pending

NullVoxPopuli:
this doesn't cover any error cases, but we've yet to have a way to capture errors, except maybe through a BuildEngineBuild record -- would need to investigate how that'd be accessed -- but that's out of scope for this PR

https://developertown.visualstudio.com/SIL%20International%20Scriptoria/_workitems/edit/34286

### DIFF
--- a/source/SIL.AppBuilder.Portal.Frontend/src/ui/components/product-artifact/index.tsx
+++ b/source/SIL.AppBuilder.Portal.Frontend/src/ui/components/product-artifact/index.tsx
@@ -27,13 +27,13 @@ export default function Builds({ product }) {
       <ResourceSelect
         items={sortedBuilds}
         labelField={(build: ProductBuildResource) => {
-          const version = attributesFor(build).version;
+          const version = attributesFor(build).version || ` ${t('projects.buildPending')} `;
 
           if (build === sortedBuilds[0]) {
             return t('projects.latestBuild', { version });
           }
 
-          return version || t('projects.buildPending');
+          return version;
         }}
         value={activeVersion}
         onChange={setActiveVersion}
--- a/source/SIL.AppBuilder.Portal.Frontend/src/ui/components/product-artifact/index.tsx
+++ b/source/SIL.AppBuilder.Portal.Frontend/src/ui/components/product-artifact/index.tsx
@@ -27,6 +27,16 @@ export default function Builds({ product }) {
       <ResourceSelect
         items={sortedBuilds}
         labelField={(build: ProductBuildResource) => {
+          // Looking at the default value here, you may notice that there are extra spaces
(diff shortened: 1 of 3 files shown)


source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/projects/show/__tests__/product-files-acceptance-test.tsx
```diff
@@ -11,6 +11,161 @@ import {
 
 import page from './page';
 
+async function visitTheFilesTab() {
```
NullVoxPopuli: this stuff here is mostly an extraction refactor

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/product-artifact/index.tsx
```diff
@@ -17,17 +17,17 @@ export default function Builds({ product }) {
     productBuilds: (q) => q.findRelatedRecords(product, 'productBuilds'),
   });
 
-  const [activeVersion, setActiveVersion] = useState((productBuilds || [])[0]);
-
-  const sortedBuilds = productBuilds.sort(compareVia((build) => attributesFor(build).version));
+  const sortedBuilds = productBuilds.sort(
+    compareVia((build) => attributesFor(build).version, true)
```
NullVoxPopuli: this just inverts the sort

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/product-artifact/index.tsx
```diff
@@ -17,17 +17,17 @@ export default function Builds({ product }) {
...
+  const [activeVersion, setActiveVersion] = useState((sortedBuilds || [])[0]);
 
   return (
     <div data-test-build>
       <ResourceSelect
         items={sortedBuilds}
         labelField={(build: ProductBuildResource) => {
-          console.log(build);
-          const version = attributesFor(build).version;
+          const version = attributesFor(build).version || ` ${t('projects.buildPending')} `;
```
abillingsley: is the whitespace before and after `${t('projects.buildPending')}` intentional?

NullVoxPopuli: it is yeah, the rendered output either be:

`Current Build (v1.0.0)`
or
`Current Build ( Build Pending )`

without the spaces it does:
`Current Build (build Pending)`
so, I guess, I should add "this" comment as an inline-comment, and then make an issue to investigate further (doing that now)

PR #597, opened 2019-04-03
Allows for better UX when interacting with the product store selection on the product show page

NullVoxPopuli:
This just adds an overlay / loader to demonstrate that something is happening

https://developertown.visualstudio.com/SIL%20International%20Scriptoria/_workitems/edit/34492


source/SIL.AppBuilder.Portal.Frontend/src/ui/components/loaders/overlay-loader.tsx
```diff
@@ -0,0 +1,25 @@
...
+import RectLoader from './rect-loader';
+
+export default function OverlayLoader(props: any) {
+  return (
+    <div
+      data-test-overlay-loader
+      className='flex-row flex-grow justify-content-conter align-items-center'
+      {...props}
+      style={{
+        opacity: 0.5,
```
NullVoxPopuli: inline styles are cool in react, right?

these aren't reusable, so here they be

NullVoxPopuli: maybe they would be if we had a legit style compiler though.

one day

source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/projects/show/overview/products/selection-manager/store-selection/index.tsx
```diff
@@ -23,53 +25,64 @@ interface INeededProps {
...
+      // loading state again
+      // setLoading(false);
+    },
+    [selected]
+  );
 
-    const { name } = attributesFor(productDefinition);
+  const { name } = attributesFor(productDefinition);
 
-    const selectProps = {
```
NullVoxPopuli: these props just moved inline

source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/projects/show/overview/products/selection-manager/store-selection/index.tsx
```diff
@@ -23,53 +25,64 @@ interface INeededProps {
...
+  return (
+    <Modal
+      data-test-project-product-store-select-modal
+      open={true}
+      className='medium p-relative'
+      style={{ zIndex: 1 }}
+      closeIcon={<CloseIcon className='close-modal' />}
+      onClose={onStoreCancel}
+    >
+      {isLoading && <OverlayLoader style={{ zIndex: 2 }} />}
```
NullVoxPopuli: this, along with the Modal's style addition is the main change in this file.

source/SIL.AppBuilder.Portal.Frontend/src/ui/components/loaders/overlay-loader.tsx
dalinarkholin: They still clutter the jsx, and are meaningless unless you are styling. Which is why I keep them in a local stylesheet as a sibling. :)

dalinarkholin: I will let it pass THIS time ;)

NullVoxPopuli: they do. This was *pure* laziness

PR #598, opened 2019-04-03
fixes a problem with the superadmin edit screen of stores

NullVoxPopuli:
- gets rid of HoCs
 - uses hooks for more predictable data lifecycles

### DIFF
--- a/source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/admin/settings/stores/-components/form.tsx
+++ b/source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/admin/settings/stores/-components/form.tsx
@@ -3,7 +3,7 @@ import { withValue } from 'react-state-helpers';
 
 import { query } from '@data';
 
-import { attributesFor, useOrbit, idFromRecordIdentity } from 'react-orbitjs';
+import { attributesFor, useOrbit, useQuery, idFromRecordIdentity } from 'react-orbitjs';
 import { Dropdown } from 'semantic-ui-react';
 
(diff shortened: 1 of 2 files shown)


source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/admin/settings/stores/-components/form.tsx
```diff
@@ -3,7 +3,7 @@ import { withValue } from 'react-state-helpers';
 
 import { query } from '@data';
```
NullVoxPopuli: this file is just converted to hooks, which uses the `useQuery` hook from dt/react-orbitjs, which has some race-condition fixes over the HoC version of query

source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/admin/settings/stores/edit.tsx
```diff
@@ -1,40 +1,47 @@
...
+
+  const {
+    result: { store },
+    isLoading,
+  } = useQuery({
     store: [(q) => buildFindRecord(q, 'store', id), buildOptions({ include: ['store-type'] })],
-  })),
-  withLoader(({ store }) => !store),
-  withProps(({ history, store }) => {
-    const { update } = useDataActions();
```
NullVoxPopuli: the root cause of the issue was that base on whether or not we were in a loading state, we'd conditionally use these hooks.
With hooks, they need to _always_ be used, because of internal implementation details (which could be avoided, imo, but whatever (WeakMap!))

PR #599, opened 2019-04-03
Save workflow comment in product transaction

davidmoore1:
Save workflow comment in product transaction


source/OptimaJet.DWKit.StarterApplication/Controllers/DataController.cs
```diff
@@ -118,7 +128,23 @@ public async Task<ActionResult> ChangeData(string name, string data)
                 {
                     throw new Exception("Access denied!");
                 }
-
+                switch (name)
```
chrisvire: Please add a block comment explaining why we are doing this and that we need to get this addressed with DWKit.

source/OptimaJet.DWKit.StarterApplication/Metadata/Forms/SIL_AppBuilders_App_Store_Review.json
```diff
@@ -6,14 +6,24 @@
...
       },
       "onSetStateClick": {
         "active": true,
         "actions": [
-          "workflowSetState"
-        ]
+          "validate",
+          "save",
+          "workflowSetState",
+          "gridRefresh"
```
chrisvire: `refresh` instead of `gridRefresh`?

PR #600, opened 2019-04-03
Update Workflow for Readiness Check, Approval Rejection, and Cleanup

chrisvire:
* Re-Add Readiness Check state to Default Workflow
* Add Approval Pending state to Default Workflow
* Add Terminated state to Default Workflow
* Reformat Readiness Check form
* Add Approval Pending form
* Simplify Published form
* Rename Project Approval title to Product Approval
* Rename App Configuration title to App Builder Configuration
* Only return actions for products that have been built.

### DIFF
--- a/scripts/DB/default_workflow.sql
+++ b/scripts/DB/default_workflow.sql
@@ -208,6 +208,12 @@ INSERT INTO "WorkflowScheme" ("Code", "Scheme") VALUES
       <Designer X="40" Y="20" />
     </Activity>
     <Activity Name="Approval Pending" State="Approval Pending" IsInitial="False" IsFinal="False" IsForSetState="True" IsAutoSchemeUpdate="True">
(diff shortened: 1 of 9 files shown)


chrisvire:
@davidmoore1 The problem was the comments added when switching to the `Approval Pending` state (using the `Hold`) command.  The comment stayed on the product and not copied to ProductTransition.  The first time from `Approval` to `Approval Pending` the comment was copied to the UserTask.  When I tried to update the comment (pressed `Hold` while starting in `Approval Pending`). it wasn't updated.

chrisvire:
@davidmoore1 I figured out why the ProductTransition was not updating.  I forgot the WriteProductTransition and UpdateProductTransition call in the workflow scheme.

davidmoore1:
Looks good

PR #603, opened 2019-04-07
Remove database trigger that is not being used

chrisvire:


chrisvire:
@davidmoore1: I put you on as a reviewer so that you knew that I was doing this work.  I needed it for the migration (seemed the database was doing a lot of extra work when I did the mass data import).  I will merge it when it passes tests.

PR #605, opened 2019-04-08
Workflow Fixes

NullVoxPopuli:
Included:
 - Cleanup from other PR
 - Workflow form advancement is restored, due to using our own SignalR Connecter, mostly copied from DWKit, so that when we connect, an exception is not thrown, and the workflow page stops working.  There still an issue where the SignalR Hub on the backend needs some auth work, because it remains either in a connecting or disconnected state. But interacting with the workflow is, in general, much nicer than it was before.
 - Use the correct connection flags so that we can ensure that the dwkit hub is connected, and not over-try-to-connect.
 - Disconnect from the hub when we navigate away from it.
 - Subclass the DWkit-provided hub so we can hack at it more easily if we need to


source/OptimaJet.DWKit.StarterApplication/Utility/DWKitHub.cs
```diff
@@ -0,0 +1,38 @@
+﻿using System;
```
NullVoxPopuli: VSCode is being broken for me atm, so I don't have an easy way to clean this up :(

source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/workflow/signalr.jsx
```diff
@@ -1,10 +1,13 @@
 import * as signalR from '@aspnet/signalr';
```
NullVoxPopuli: this is mostly copied from DWKit's source. modified a bit to be less verbose in the logging, and then also auth via jwt

NullVoxPopuli:
If the user tasks weren't updating, then we have an issue with websockets on the deployed environment or something. I'll take a stab

source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/tasks/display.tsx
```diff
@@ -10,7 +10,6 @@ import { useLiveData } from '~/data/live';
 export default function Display({ userTasks }: IUserTaskListProps) {
   const { t } = useTranslations();
   useLiveData('products');
-  useLiveData('user-tasks');
```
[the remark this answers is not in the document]
NullVoxPopuli: It was redundant

PR #621, opened 2019-04-22
Update ReassignUserTasksForProduct to account for product without workflow

chrisvire:
The old model was that all products would always have a running workflow and rebuild and republish would be transitions in the same workflow.  The new model is that the primary workflow will run to get the app published the first time and then complete (and the workflow will stop).  Rebuild and Republish are separate smaller workflows.  

So when a project is reassigned to another user, the ReassignUserTasksForProduct has to account for the situation where there is currently no running workflow for the product and not do any work.


chrisvire:
This is for https://developertown.visualstudio.com/SIL%20International%20Scriptoria/_workitems/edit/34739

PR #629, opened 2019-05-03
Add region to token request response

davidmoore1:


chrisvire:
Should wait for BuildEngine to deploy before merging

PR #630, opened 2019-05-07
Change state names and fix republish

davidmoore1:
* Add env parameter to republish action parameter
* Change state names to match activity names
* Fix case where executed transition doesn't match the actual transition

Apply after Build Engine PR # 19


source/OptimaJet.DWKit.StarterApplication/Services/Workflow/WorkflowActivityMonitorService.cs
```diff
@@ -56,7 +56,7 @@ private void ActivityChanged(ProcessActivityChangedEventArgs args, WorkflowRunti
 
             Log.Information($":::::::::: ActivityChanged: pid={args.ProcessId.ToString()}, scheme={args.SchemeCode}, activity={args.CurrentActivityName}, state={args.CurrentState}, last={args.PreviousState}");
             WorkflowProductService.TransitionType transitionType = WorkflowProductService.TransitionType.Other;
-            if (args.ExecutedTransition != null && (args.CurrentActivityName != args.PreviousActivityName))
```
chrisvire: I think this should be pulled out into a separate method GetTransitionType with a comment to explain why we are checking the states.  Something like:
```
// In DWKit 2.4.1, when executing a timeout transition back to the same state,
// args.ExecutedTransition isn't the transition that is actually being executed 
// (seems to be the first transition of all defined).  In this case, ignore the
// transition type.
```

PR #651, opened 2019-05-22
WIP: Change product task information

davidmoore1:
Display product task list with names of those it is waiting on or 'Scriptoria' if waiting for an automated activity.


davidmoore1:
The purpose of this is to change the task line to always display on the project detail if there are tasks being performed with the list of allowed users from the product transition and to display the continue button if a task for the current user is defined for the project.

If the one or more users are in the list and the current
Screen Shot 2019-06-03 at 10 26 54 AM
 user has a task, the following is displayed:

PR #666, opened 2019-06-03
Feature/change project product task list

davidmoore1:
Change the current project detail product task line.
Display the list of allowed users in place of the word "you" and always display the task line if open product transitions for the product exist.

Display the user name and the continue button if the current user has a user task for the task being displayed: 
Screen Shot 2019-06-03 at 10 26 54 AM

Display the allowed user list without the continue button if the current user does not have a task:
Screen Shot 2019-06-03 at 10 26 54 AM

Show waiting on scriptoria if the allowed names list is empty
Screen Shot 2019-06-03 at 10 28 20 AM

Display nothing if publish is completed
Screen Shot 2019-06-03 at 10 28 49 AM


davidmoore1:
Put the wrong screen shot up for if current user doesn't have a task:
Screen Shot 2019-06-03 at 10 27 24 AM

source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/projects/show/overview/products/item/tasks.tsx
```diff
@@ -13,70 +13,137 @@ import { useCurrentUser } from '~/data/containers/with-current-user';
...
+
 interface IProps {
   product: ProductResource;
 }
-
 export default function ProductTasksForCurrentUser({ product }: IProps) {
   const { t } = useTranslations();
   const { currentUser } = useCurrentUser();
   const { dataStore } = useOrbit();
+  const productRemoteId = remoteIdentityFrom(dataStore, product).keys.remoteId;
```
NullVoxPopuli: I think there is a method that'll straight up just give you the remote id

source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/projects/show/overview/products/item/tasks.tsx
```diff
@@ -13,70 +13,137 @@ import { useCurrentUser } from '~/data/containers/with-current-user';
...
-  } catch (e) {
-    console.debug(
-      'error occurred',
-      e,
-      'we probably need to PR to orbit.js, as this is a race condition scenario that causes the error',
-      'variables that are in use that may be related to this error: ',
-      ['product: ', product, 'currentUser: ', currentUser]
-    );
-  }
+  useEffect(() => {
```
NullVoxPopuli: instead of using useEffect, I think you can define a callback with useCallback and pass that to AsyncWaiter (search around for other usages), it'll also give you loading and error states :)

source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/projects/show/overview/products/item/tasks.tsx
```diff
@@ -13,70 +13,137 @@ import { useCurrentUser } from '~/data/containers/with-current-user';
...
+      } catch (e) {
+        setFetchComplete(true);
+      }
+    }
+    fetcher();
+  }, [productRemoteId]);
 
   const { relativeTimeAgo } = useTimezoneFormatters();
   const { pathToWorkflow } = useUserTaskHelpers();
+  const getCurrentUserTask = () => {
```
NullVoxPopuli: this probably needs to be extracted to its own hook -- also remember to use useCallback :)

source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/projects/show/overview/products/item/tasks.tsx
```diff
@@ -13,70 +13,137 @@ import { useCurrentUser } from '~/data/containers/with-current-user';
...
+      console.debug(
+        'error occurred',
+        e,
+        'we probably need to PR to orbit.js, as this is a race condition scenario that causes the error',
+        'variables that are in use that may be related to this error: ',
+        ['product: ', product, 'currentUser: ', currentUser]
+      );
+    }
+  };
+  const formatTaskLine = (transition) => {
```
NullVoxPopuli: this should probably be a utility function as it doesn't seem to have any bearing on the component

source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/projects/show/overview/products/item/tasks.tsx
```diff
@@ -13,70 +13,137 @@ import { useCurrentUser } from '~/data/containers/with-current-user';
...
+    } else {
+      return (
+        <div key={transition.id}>
+          <span className='red-text'>{t('tasks.waiting', { waitTime })}</span>&nbsp;
+          {t('tasks.forNames', { allowedNames, activityName })}
+        </div>
+      );
+    }
+  };
+  const [currentUserTask, task] = useMemo(getCurrentUserTask, [product, currentUser]);
```
NullVoxPopuli: can't use hooks after a return (as on 116 ash 126)

source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/projects/show/overview/products/item/tasks.tsx
```diff
@@ -13,70 +13,137 @@ import { useCurrentUser } from '~/data/containers/with-current-user';
...
+  if (fetchComplete == false) {
```
NullVoxPopuli: this block won't be needed with the AsyncWaiter :)

source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/projects/show/overview/products/item/tasks.tsx
```diff
@@ -13,70 +13,137 @@ import { useCurrentUser } from '~/data/containers/with-current-user';
...
+      if (!taskProduct) return 'Task has no product!';
+    } else {
+      const dateTransition = attributes['date-started'];
+      if (dateTransition) {
+        waitTime = relativeTimeAgo(dateTransition);
+      }
+    }
+    if (currentUserTask) {
+      return (
+        <div key={transition.id}>
```
NullVoxPopuli: components could probably be extracted here

source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/projects/show/overview/products/item/tasks.tsx
NullVoxPopuli: react's hook implementation requires that all hooks be used every time the component is rendered. so sometimes having an additional hook (or fewer hooks) throws off the hook manager

source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/projects/show/overview/products/item/tasks.tsx
davidmoore1: I moved it back inline and redid the return to format the line in a more conventional way.

davidmoore1:
The current commit generally works but the task line should automatically update if the user is waiting on the project detail screen and it doesn't.  Current example: I have two users in two browsers viewing a project detail, org admin and project owner.  Project is waiting for task to begin build (which will legitimately fail).  The owner screen shows the "Continue" button, indicating it is waiting for a task.  Both screens show the Current transition and tasks array.  From a third browser, the task is started, so there is no longer a user task.  Both screens refresh, showing an empty task array, but transition does not update.  Continue button disappears.  Task runs and fails.  A notification is issued.  On the user screen, the continue button reappears and the time updates.    The transition still has not updated.  On the org admin screen, even though the query of tasks is run, it sees an empty task set is retrieved, even though this is the same query being performed on the other screen.  The time is not updated.  As with the other screen, even though the useTransitions is called, there is no change.  Reloading both screens causes the correct transitions to be seen and the correct tasks to be viewed.

source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/projects/show/overview/products/item/tasks.tsx
davidmoore1: Done

source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/projects/show/overview/products/item/tasks.tsx
davidmoore1: Removed the block

source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/projects/show/overview/products/item/tasks.tsx
davidmoore1: Extracted into separate hook using useCallback

davidmoore1:
Documenting the known issue with this.  If you leave the project detail screen up, it will initially show the task that is being worked on and the correct associated person.  However, if the task changes, it will not update except by leaving and reentering the project detail page or by refreshing the page.  The continue button will update correctly and will disappear or appear depending upon whether the task is assigned to the current user.  However, the list of users and the task will only be correct when the page is initially entered.

Chris and I have discussed this and decided the additional information and the fact that is correct whenever a user enters the detail page to view it makes it more valuable than the current functionality, so we will accept the deficiency for now and hopefully come back and correct this at a later date.

PR #667, opened 2019-06-03
Fix project link on task page doesn't work

davidmoore1:
Removed invalid parameter from call to idFromRecordIdentity
Also removed red lines in code display by casting to any
as is done elsewhere in similar situations.


chrisvire:
That was all that was needed?  Great!

PR #672, opened 2019-06-07
Add status line for publish results associated with the build

davidmoore1:


source/SIL.AppBuilder.Portal.Frontend/src/ui/components/product-artifact/artifacts.tsx
```diff
@@ -26,7 +26,7 @@ export default function ProductArtifact({ product, productBuild }) {
   }
 
   return (
-    <div data-test-product-artifacts className='product-artifact w-100 m-b-lg'>
```
chrisvire: So what was `m-b-lg`?

davidmoore1: This was a large margin on the bottom of product artifacts on what used to be the last item on the screen.  When I added the status line, it added a large gap between the two divs.  I removed this so the Publication section is snugged up against the bottom and the new Div, which is there whether it contains a status line or not (just like the artifacts div is empty if no builds are present), has the large margin, so the effect is the same.

PR #681, opened 2019-06-25
Add project transition details

davidmoore1:
Updated code without tests


chrisvire:
@davidmoore1 I am going to merge so that we can include this in production.  Please continue (on another branch) to add the tests.

PR #684, opened 2019-06-26
Upgrade DotNet 2.2.0 & EntityFrameworkCore 2.2.4

chrisvire:


chrisvire:
The reason for upgrading to DotNet 2.2.0 was to allow update SignalR.  But newer Signal doesn't solve problems (and has duplicate message in tasks to return).  We need to update implementation of SignalR in /tasks and /projects/x for the notification to signal a model refresh instead of trying to pass data.

PR #694, opened 2019-07-09
Add start, end, and cancel entries to details report

davidmoore1:
Added workflow type to transition for start, stop and cancel type records
Made some minor changes to unrelated tests to eliminate warnings.


davidmoore1:
Screen Shot 2019-07-09 at 11 00 49 AM
Added new headers in bold to mark start and end of workflow sequences

PR #721, opened 2019-10-04
New project group list is inconsistent

davidmoore1:
Issues with accessing the cache for group.  This retrieves the
group information associated with the organization with a query.
All the subsequent accesses to the cache for that request then
retrieve the correct information.

### DIFF
--- a/source/SIL.AppBuilder.Portal.Frontend/src/ui/components/inputs/group-select/with-data.tsx
+++ b/source/SIL.AppBuilder.Portal.Frontend/src/ui/components/inputs/group-select/with-data.tsx
@@ -1,6 +1,6 @@
-import { useOrbit } from 'react-orbitjs';
+import { useOrbit, useQuery } from 'react-orbitjs';
 
-import { recordsWithIdIn, isRelatedTo } from '@data';
+import { recordsWithIdIn, isRelatedTo, buildOptions } from '@data';
 
 import { OrganizationResource } from '@data/models/organization';
 import { GroupResource } from '@data/models/group';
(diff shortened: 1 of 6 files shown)


source/SIL.AppBuilder.Portal.Frontend/src/ui/components/inputs/group-select/with-data.tsx
```diff
@@ -53,7 +61,8 @@ export function useScopeGroupData({
         return true;
       }
 
-      return isRelatedTo(group, 'owner', scopeToOrganization.id);
+      const retVal = isRelatedTo(group, 'owner', scopeToOrganization.id);
+      return retVal;
```
chrisvire: Why was this changed? Spurious while testing?

davidmoore1:
Additional information on original problem:  
Scenario: Logging in as super user in this case, which should allow me to create project with any defined group for the organization.  I already have one project defined for the Asia group.  

Screen Shot 2019-10-16 at 10 54 52 AM

I log in as the user and immediately go to create new project.  If you look at the list of available groups, it is correct:

Screen Shot 2019-10-16 at 10 54 13 AM

Now I exit and do something else.  The something else can be looking at the details of the existing project.  It can be going to Users and back.  It can be going to Organizational settings.  I come back and try to create a new project.  The list is now limited to the groups associated with the projects in the project list (in this case, one Asia project, so Asia).  The query to the cache for the list of all groups returns only the one group.

Screen Shot 2019-10-16 at 10 56 54 AM

chrisvire:
@davidmoore1 I am going to merge. I don't know of a better way to fix this right now. Maybe we will figure it out when SignalR/Orbit work is worked on.

PR #723, opened 2019-10-11
Problems with accepting user invitation

davidmoore1:
Main obstacle was that fetcher did not initialize the fetch count as
a number so incrementing it didn't actually do anything.
Now proceeds to tasks on completion


davidmoore1:
Also modified class to current functional approach

PR #745, opened 2019-10-16
Display error on not active user login

davidmoore1:


davidmoore1:
Now displays the following screen when inactive user logs in
Screen Shot 2019-11-12 at 2 01 24 PM

PR #750, opened 2019-11-11
Changes to move user settings from popups to screens

davidmoore1:
This first change moved the profile and the group settings
It has not removed any of the earlier code at this point


davidmoore1:
Overall note, after all is said and done, a lot of the code is just moved around.  Some modules in multi-select were changed name to multi-list since no selection was being done there anymore, so from the github point of view, several new modules were added and deleted.  Code that was in the dropdowns was moved into tabs in the new editsettings code.  That said, a number of changes were made to get them to work in the new setup. Tests had to be substantially changed but are essentially testing the same points.

PR #751, opened 2019-11-19
Implement Project Import

chrisvire:
- [x] Backend Datamodel
- [x] Frontend Datamodel
- [x] Frontend UI
- [x] Backend Processing of Import
- [x] Backend Email Results
- [x] Document JSON Payload and link in Page

1. Add `Import Project` on `My Projects` page.

image

2. Prompt for Group, Application Type, and JSON File of Projects and Products.

image

3. Notify User they will receive an email with results

image

4. Process Project Import.  There should be projects created, products created, and workflow started for the projects.

image

image

5. Report to user in email on the results.

image


chrisvire:
I believe that `LGTM analysis: C#` is failing since it is using .NET Core 3.1.

PR #752, opened 2019-11-25
Changes to clean up LGTM check

davidmoore1:


source/SIL.AppBuilder.Portal.Frontend/src/data/containers/with-relationship.tsx
```diff
@@ -94,11 +93,7 @@ function retrieveManyToMany(
   });
 }
 
-async function retriveDirectRelationship(
-  dataStore: Store,
-  sourceModel: ResourceObject,
-  relationshipName: string
-) {
+async function retriveDirectRelationship() {
   // TODO: add detection for hasOne vs hasMany, via lookup of the schema from dataStore
```
chrisvire: @davidmoore1: Does this concern you too? Is this used a lot? What are the conditions when this not implemented code could be hit?

Maybe these should be commented out instead of deleted? I assume that this was stubbed out to be done in the future. We might need to know what parameters would be needed when eventually implemented.

chrisvire: I added Trello Card: https://trello.com/c/DDmOwtEB/146-research-retrievedirectrelationship

PR #766, opened 2020-01-10
Separate add/remove product button into two separate buttons

davidmoore1:
This pull request separates the add remove product button into two separate add and remove buttons.  Pressing add shows only the products to be added which are removed from the list as they are added.  Pressing remove shows only the existing products and removes them from the list as they are deselected.  This PR also fixes issues where the underlying list on the project screen was not updated when products were added and deleted.
Project screen:
Screen Shot 2020-01-10 at 10 46 12 AM
Remove product popup:
Screen Shot 2020-01-10 at 10 46 51 AM
Add product popup:
Screen Shot 2020-01-10 at 10 46 40 AM

### DIFF
--- a/source/SIL.AppBuilder.Portal.Frontend/src/data/containers/with-relationship.tsx
+++ b/source/SIL.AppBuilder.Portal.Frontend/src/data/containers/with-relationship.tsx
@@ -117,7 +117,10 @@ export function relationsFromPath(
 
   for (let i = 0; i < relationshipPath.length; i++) {
     segment = relationshipPath[i];
-    assert(`can't access segment (${segment}) on falsey model.`, !!model);
+    if (!model) {
+      console.debug(`can't access segment (${segment}) on falsey model.`);
+      return models;
+    }
 
     model = dataStore.cache.query((q) => q.findRelatedRecord(model, segment));
     models.push(model);
--- a/source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/projects/show/__tests__/products-with-stores-acceptance-test.tsx
+++ b/source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/projects/show/__tests__/products-with-stores-acceptance-test.tsx
@@ -12,6 +12,8 @@ import { MultiSelectInteractor } from '~/ui/components/inputs/multi-select/-page
 
 import { ProductsInteractor } from '~/ui/routes/projects/show/overview/products/-page.ts';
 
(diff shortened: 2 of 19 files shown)


source/SIL.AppBuilder.Portal.Frontend/src/data/containers/with-relationship.tsx
```diff
@@ -117,7 +117,10 @@ export function relationsFromPath(
 
   for (let i = 0; i < relationshipPath.length; i++) {
     segment = relationshipPath[i];
-    assert(`can't access segment (${segment}) on falsey model.`, !!model);
+    if (!model) {
+      console.debug(`can't access segment (${segment}) on falsey model.`);
+      return models;
+    }
```
chrisvire: What is this about? Did you intend to leave it this way?

source/SIL.AppBuilder.Portal.Frontend/src/ui/routes/projects/show/__tests__/products-with-stores-acceptance-test.tsx
```diff
@@ -282,15 +286,18 @@ describe('Acceptance | Project View | Products With Stores', () => {
             await when(() =>
               assert(stores.isOverlayLoaderVisible, 'expected overlay loader to briefly be visible')
             );
-            await when(() => products.checkedItems().length > 0);
+            await when(() => products.itemNamed('android_s3').isVisible);
```
chrisvire: Is it OK to use this specific name in the test?

chrisvire:
I talked with Bill Dyck about the proposed UI changes.  I want to be careful to not be [bikeshedding](https://en.wiktionary.org/wiki/bikeshedding).  

Here is our feedback and some proposed changes:
1. the checkboxes are not needed anymore and are actually confusing
2. the titles of the dialogs should reflect what we want the user to do:
    a. Click Product to Add
    b. Click Product to Remove
3. the gray Android icon in the Add dialog seems like it is disabled/grayed out
4. the X and the done button are redundant (and the X and the cancel button when setting the store).  Should we do anything? Maybe not. Seeing an X and  Done button makes it seem like the actions happen when you click Done and you can cancel the dialog without having any affect.

davidmoore1:
Dio what do you want in place of checkboxes

chrisvire:
> Dio what do you want in place of checkboxes

I think we would suggest to remove the checkbox from the entry.

PR #851, opened 2020-09-03
Change save button texts in admin to "Save"

davidmoore1:


chrisvire (commented):
What about having "Save" and "Add" buttons?
```
{organization
   ? t('common.save')
   : t('common.add')}
```

PR #858, opened 2020-09-18
Allow editing of user name

davidmoore1:
Also use full user name in title of edit profile, title of edit user settings, and the list of users.


source/SIL.AppBuilder.Portal.Frontend/src/translations/locales/en-us.json
```diff
@@ -159,6 +159,7 @@
     "uploadPicture": "Upload new picture",
     "firstName": "First name",
     "lastName": "Last name",
+    "name": "Name",
```
chrisvire, line 162: Should this be "Display Name"?

chrisvire (changes requested):

chrisvire (approved):
Thanks

PR #861, opened 2020-10-08
Create build or release error message with correct build engine URL

davidmoore1:
when using default build engine

### DIFF
--- a/source/OptimaJet.DWKit.StarterApplication/Services/BuildEngine/BuildEngineBuildService.cs
+++ b/source/OptimaJet.DWKit.StarterApplication/Services/BuildEngine/BuildEngineBuildService.cs
@@ -272,7 +272,9 @@ namespace OptimaJet.DWKit.StarterApplication.Services.BuildEngine
         {
             ClearRecurringJob(product.Id);
             await SaveArtifacts(product, buildEngineBuild, false);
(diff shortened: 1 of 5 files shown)


source/OptimaJet.DWKit.StarterApplication/Services/BuildEngine/BuildEngineBuildService.cs
```diff
@@ -272,7 +272,9 @@ protected async Task BuildCreationFailedAsync(Product product, BuildResponse bui
         {
             ClearRecurringJob(product.Id);
             await SaveArtifacts(product, buildEngineBuild, false);
-            var buildEngineUrl = product.Project.Organization.BuildEngineUrl + "/build-admin/view?id=" + product.WorkflowBuildId.ToString();
+            var buildEngineEndpoint = GetBuildEngineEndpoint(product.Project.Organization);
```
chrisvire, line 275: I assume that `GetBuildEngineEndpoint` get the default BuildEngine if the Organization is using the default.

chrisvire (approved):
👍

source/OptimaJet.DWKit.StarterApplication/Services/BuildEngine/BuildEngineBuildService.cs
davidmoore1: Correct

PR #862, opened 2020-10-16
Timezone dropdown should appear over notification options

davidmoore1:
Also fix problem with user settings arrow icon not appearing on the same line as the timezone select box


chrisvire (approved):
👍

PR #881, opened 2020-12-03
Automatically refresh product tasks when Orbit cache updates

garrett-hopper:
scrot

I'm not sure what's causing this error on the tasks screen. (It's not caused by this change.)
scrot


chrisvire:
👍

PR #883, opened 2020-12-04
Subscribe to product transitions instead of user tasks

garrett-hopper:
Subscribing to the transitions ensures we're always showing the correct transition, however there are still some unnecessary refreshes when a new transition comes through that doesn't actually update the display. The `continue` button also needs to be fixed.
scrot


garrett-hopper:
This should now work as intended. The correct active product transition should always be displayed, and the `continue` button should always appear if the user has a task for the product.
scrot

chrisvire:
I will merge this and test it.

chrisvire:
👍

PR #884, opened 2020-12-07
Fix error when rendering task without product

garrett-hopper:
Prevent the task row from rendering if the task is missing a product
in order to avoid `Cannot read property 'type' of null` error being
thrown by Orbit queries.

It looks like a task removal event isn't being sent by the server in addition to the product removal event, and it's nontrivial to cascade the product removal to remove related tasks as well.

scrot

Previous error:
image


garrett-hopper:
Fixed by #885

PR #889, opened 2020-12-09
Switch from TravisCI to GitHub actions

chrisvire:


chrisvire:
The "failed" "build (push)" was that I cancelled the build since it didn't make sense to have two running.

PR #894, opened 2020-12-23
Fix product transition task infinite loading

garrett-hopper:
Previously the loading animation would play forever if no transition could be found. It will only display initially now.


chrisvire:
Let's give it a try...

PR #895, opened 2020-12-23
Live update product actions

garrett-hopper:
Product actions will be refetched when the product transitions change.


chrisvire:
👍

PR #897, opened 2021-01-24
fix error when product no longer exists

garrett-hopper:
This PR resolves errors (`Cannot read property 'type' of null` specifically) when a product is deleted from the database without being removed from the Orbit cache. This was tested manually be deleting the product from the `Products` table manually, however I believe it occurs in the wild due to websocket failures which result in the product removal not being sent to the client.

Due to a bug in OrbitJS, the project display screen would request the project as well as its associated products. Because the product no longer exists, it would update (not delete) the product entity which no longer exists in the database to have a relationship to a project of `null` which would result in the error. This is bypassed filtering the products outside of Orbit.

Once the product has been deleted, there are several requests that will 404 and fail to parse as JSON. These 404 responses will now gracefully fail without a JSON parse error.


chrisvire (approved):
LGTM

PR #899, opened 2021-02-23
Fix accept org membership invite

chrisvire:
* Change backend to only process one patch (execute in background job)
* Change API to return Invite instead of OrganizationMemembership
* Change frontend to redirect to Organization Membership Changed message

image (3)


davidmoore1 (approved):
Looks very good

PR #967, opened 2022-04-19
Change notification handling

davidmoore1:
Always send superuser notification emails immediately
Only add notification to table if email is to be sent
Change default email send delay from 60 minutes to 1


chrisvire (approved):
LGTM!

PR #971, opened 2022-06-02
Add default timeout for remote queries

davidmoore1:


chrisvire:
@davidmoore1 Thats it?

chrisvire (approved):

PR #1005, opened 2023-11-01
Investigation into Svelte re-write

chrisvire:


source/SIL.AppBuilder.Portal/package.json
```diff
@@ -51,5 +51,8 @@
 	"type": "module",
 	"volta": {
 		"node": "18.16.1"
+	},
+	"dependencies": {
+		"zod": "^3.23.8"
```
chrisvire, line 56: Normally, all dependencies with Svelte-kit are devDependencies.

source/SIL.AppBuilder.Portal/src/lib/components/InternationalizedDataBox.svelte
```diff
@@ -4,19 +4,31 @@
     
 -->
 <script lang="ts">
```
chrisvire, line 6: FYI, "Internationalized" is probably not needed in the component name.  Everything should internationalized.

PR #1019, opened 2024-09-24
Workflows with XState

FyreByrd:
Uses the XState library to create FSM instances that model the desired workflow behavior.

Features:
- Persistence: Each instance has an associated DB record (in `WorkflowInstances`) that stores enough information to restore to the current state.
- User Tasks: Each instance updates the `UserTasks` table with the appropriate information.
- Product Transitions: Each instance updates the `ProductTransitions` table with the appropriate information.
- Tasks: The appropriate actions are provided based on the user in the `/tasks/[product_id]` and are handled by the FSM.
- Admin: A barebones table of instances is provided in `/admin/workflows`. 
    - Clicking on an entry will take a user to a visualization in `/admin/workflows/[product_id]`
    - While in the visualization, a user can select a state node and then press a button in a menu to jump the machine state to the selected state

Future Priorities:
- Hook actions into backend
- Authentication
- Republish and Rebuild workflows
- Workflow instance filtering/sorting in `/admin/workflows`
- Improve visualization (and visualization efficiency)


source/SIL.AppBuilder.Portal/package.json
```diff
@@ -34,6 +34,7 @@
     "@typescript-eslint/eslint-plugin": "^6.0.0",
     "@typescript-eslint/parser": "^6.0.0",
     "@vvo/tzdb": "^6.142.0",
+    "@xstate/svelte": "^3.0.4",
```
FyreByrd, line 37: Remove. Not used

source/SIL.AppBuilder.Portal/package.json
```diff
@@ -66,7 +67,9 @@
     "@auth/express": "^0.6.0",
     "@auth/sveltekit": "^1.4.2",
     "@bull-board/express": "^5.21.3",
+    "@statelyai/inspect": "^0.4.0",
```
FyreByrd, line 70: Remove. Not used

source/SIL.AppBuilder.Portal/common/workflow/db.ts
FyreByrd, line 1: Add explanatory comments

source/SIL.AppBuilder.Portal/src/lib/filterObject.ts
```diff
@@ -0,0 +1,14 @@
+//source: https://www.steveruiz.me/posts/how-to-filter-an-object
+
+type Entry<T> = {
+  [K in keyof T]: [K, T[K]]
+}[keyof T]
+
+export function filterObject<T extends object>(
```
FyreByrd, line 7: Comment?

source/SIL.AppBuilder.Portal/src/routes/(authenticated)/admin/workflows/[product_id]/+page.server.ts
```diff
@@ -0,0 +1,86 @@
...
+            }
+          },
+          ProductDefinition: {
+            select: {
+              Name: true
+            }
+          },
+          ProductTransitions: {
+            select: {
+              DateTransition: true,
```
Comment (FyreByrd) [inline source/SIL.AppBuilder.Portal/src/routes/(authenticated)/admin/workflows/[product_id]/+page.server.ts:39]: Null DateTransition?

source/SIL.AppBuilder.Portal/src/routes/(authenticated)/admin/workflows/[product_id]/+page.server.ts
```diff
@@ -0,0 +1,86 @@
...
+import { fail, superValidate } from 'sveltekit-superforms';
+import { valibot } from 'sveltekit-superforms/adapters';
+import * as v from 'valibot';
+
+const jumpStateSchema = v.object({
+  state: v.string()
+});
+
+export const load: PageServerLoad = async ({ params, url, locals }) => {
+  const actor = createActor(DefaultWorkflow, {
```
Comment (FyreByrd) [inline source/SIL.AppBuilder.Portal/src/routes/(authenticated)/admin/workflows/[product_id]/+page.server.ts:14]: Probably don't need to create actor here. Can just use Value and Context in load.

source/SIL.AppBuilder.Portal/src/routes/(authenticated)/admin/workflows/[product_id]/+page.svelte
```diff
@@ -0,0 +1,223 @@
...
+    position: absolute;
+    z-index: 5; /* position above canvas, but below drawer */
+    right: 0;
+  }
+
+  details > summary {
+    display: block; /* remove arrow */
+  }
+
+  details:not([open]) > summary strong {
```
Comment (FyreByrd) [inline source/SIL.AppBuilder.Portal/src/routes/(authenticated)/admin/workflows/[product_id]/+page.svelte:214]: What is this doing?

source/SIL.AppBuilder.Portal/src/routes/(authenticated)/admin/workflows/[product_id]/+page.svelte
```diff
@@ -0,0 +1,223 @@
...
+      () => {}, // clear
+      () => {}, // drawEdge
+      (node: Springy.Node, position: Springy.Physics.Vector) => {
+        // drawNode
+        positions[node.id] = position;
+      },
+      () => {}, // onRenderStop
+      () => {}, // onRenderStart
+      () => {
+        ready = true;
```
Comment (FyreByrd) [inline source/SIL.AppBuilder.Portal/src/routes/(authenticated)/admin/workflows/[product_id]/+page.svelte:66]: Do we really want this visible this soon?

source/SIL.AppBuilder.Portal/src/routes/(authenticated)/tasks/[product_id]/+page.server.ts
```diff
@@ -1,20 +1,43 @@
...
+  productDescription?: string; //Product.ProductDefinition.Description
+  appType?: string; //Product.ProductDefinition.ApplicationTypes.Description
   projectLanguageCode?: string; //Product.Project.Language
-}
+};
 
 export const load = (async ({ params, url, locals }) => {
+  const session = await locals.auth();
+  // TODO: permission check
+  const actor = createActor(DefaultWorkflow, {
```
Comment (FyreByrd) [inline source/SIL.AppBuilder.Portal/src/routes/(authenticated)/tasks/[product_id]/+page.server.ts:35]: Probably don't actually need the actor. Could just use snapshot

source/SIL.AppBuilder.Portal/src/routes/(authenticated)/tasks/[product_id]/+page.svelte
```diff
@@ -7,20 +7,21 @@
 </script>
 
 <div class="p-5">
-  <form>
+  <form method="POST">
     {#if data.actions?.length}
     <div class="flex flex-row gap-x-3">
       {#each data.actions as action}
-      <input type="submit" class="btn" value={action}>
+      <input type="submit" name="action" class="btn" value={action}>
```
Comment (FyreByrd) [inline source/SIL.AppBuilder.Portal/src/routes/(authenticated)/tasks/[product_id]/+page.svelte:14]: Change to flowAction. Potential name conflicts with form.action

source/SIL.AppBuilder.Portal/common/workflow/default-workflow.ts
```diff
@@ -0,0 +1,948 @@
...
+      createSnapshot(context.currentState, context);
+      updateUserTasks(
+        context.productId,
+        roles ? (Array.isArray(roles) ? roles : [roles]) : [],
+        context.currentState,
+        event.comment
+      );
+    },
+    transit: ({ context, event }, params?: { target?: StateName }) => {
+      const command = event.type.split(':');
```
FyreByrd, line 63: Fix this

source/SIL.AppBuilder.Portal/common/workflow/default-workflow.ts
```diff
@@ -0,0 +1,948 @@
...
+          update: {},
+          create: {
+            Snapshot: '',
+            ProductId: context.productId
+          }
+        });
+      },
+      always: [
+        {
+          guard: { type: 'canJump', params: { target: 'Readiness Check', level: AdminLevel.High } },
```
FyreByrd, line 126: What if this is correct start?

source/SIL.AppBuilder.Portal/src/routes/(authenticated)/tasks/[product_id]/+page.server.ts
```diff
@@ -61,47 +98,107 @@ export const load = (async ({ params, url, locals }) => {
...
       projectName: product?.Project.Name,
-      projectDescription: product?.Project.Description,
-      storeDescription: product?.Store?.Description,
-      listingLanguageCode: product?.StoreLanguage?.Name,
-      projectURL: product?.Project.WorkflowAppProjectUrl,
-      productDescription: product?.ProductDefinition.Name,
-      appType: product?.ProductDefinition.ApplicationTypes.Description,
-      projectLanguageCode: product?.Project.Language
+      projectDescription: product?.Project.Description
     } as Fields,
```
Comment (7dev7urandom) [inline source/SIL.AppBuilder.Portal/src/routes/(authenticated)/tasks/[product_id]/+page.server.ts:173]: Is this the only place you're using your filterObject function? Often if you're using a filterObject function it means something could be done differently. I'm not entirely opposed to having it, but think about it...

7dev7urandom (changes requested):
Looks pretty good; well done. 

Here's a first-look review. I haven't reviewed the `default-workflow.ts` stuff at all, and I haven't tested anything, but here's some comments I have looking through it rather quickly. Some questions are probably somewhat ignorant because I'm not entirely clear what your goal is, but I think they'll be helpful. 

One larger thing: you're handling the state machines with a lot of functional programming. You have a lot of functions that take `WorkflowMachine` or `WorkflowContext` as (often the first) argument, which makes me wonder if it could be done in a more object-oriented way, making things more intuitive to write.

source/SIL.AppBuilder.Portal/common/public/workflow.ts
```diff
@@ -0,0 +1,254 @@
...
+}
+
+export enum AdminLevel {
+  /** NoAdmin/OwnerAdmin */
+  None = 0,
+  /** LowAdmin */
+  Low,
+  /** Approval required */
+  High
+}
```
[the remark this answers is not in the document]
FyreByrd: By 'switch the names around' do you mean just reorder them? Or changing the meaning? Because I'm pretty sure we agree on the meaning?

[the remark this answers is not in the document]
FyreByrd: I'm not sure how to make the name more descriptive without making all usage of it unnecessarily verbose. I can add more comments?

source/SIL.AppBuilder.Portal/common/public/workflow.ts
```diff
@@ -0,0 +1,254 @@
...
+  productId?: string;
+  adminLevel?: AdminLevel;
+  environment?: BuildEnv;
+  productType?: ProductType;
+};
+
+/** Used for filtering based on AdminLevel and/or ProductType */
+export type MetaFilter = {
+  level?: AdminLevel | AdminLevel[];
+  product?: ProductType | ProductType[];
```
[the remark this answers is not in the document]
FyreByrd: I did it this way, because in most cases I am only filtering on one product type or admin level. There are a *few* instances where it was necessary to filter on multiple types. I mostly just didn't want to create unnecessary arrays when there's only one element to filter on.

source/SIL.AppBuilder.Portal/common/databaseProxy/Products.ts
```diff
@@ -1,4 +1,70 @@
...
+  // No additional verification steps
+
+  try {
+    await prisma.products.update({
+      where: {
+        Id: id
+      },
+      data: productData
+    });
+    // later: Are there any other updates that need to be done?
```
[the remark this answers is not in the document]
FyreByrd: Agreed. Standardization is good

source/SIL.AppBuilder.Portal/common/public/workflow.ts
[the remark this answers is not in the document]
7dev7urandom: I mean, is this enum used as a minimum required admin level for a user to complete an action, or as the level of admin that a user/action has, or something else? My understanding is the former, but the name implies to me the latter. The comments seem backwards to me. What does the comment "NoAdmin/OwnerAdmin" mean? That the action was performed by an admin? That the action should not require admin in any way, as indicated by the name "None"?

Does that make sense? I'd rather rename the enum to "RequiredAdminLevel" or something more descriptive like that. At least clarify the comments.

source/SIL.AppBuilder.Portal/common/public/workflow.ts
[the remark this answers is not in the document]
FyreByrd: Actually, I will go ahead and fix that everywhere that I did that in my PR, not just the meta filtering

source/SIL.AppBuilder.Portal/common/public/workflow.ts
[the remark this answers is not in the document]
FyreByrd: The way I am using the enum, I am including a state or transition in a workflow *only* if the specified admin level of the workflow type (i.e. OwnerAdmin, LowAdmin, NoAdmin, Default, etc.) matches any of the specified admin level types in the meta (or if there is no specified level).

For Example:
A workflow will only include the `Readiness Check`, `Approval`, `Approval Pending`, and `Terminated States` if the workflow's admin level is `AdminLevel.High`, which is the case, for example, in the `SIL_Default_AppBuilders_Android_GooglePlay` scheme in DWKit. 

If a workflow's admin level is `AdminLevel.Low`, it will not include the states for the approval process, but there are some actions (i.e. transitions) that require an OrgAdmin.

If a workflow's admin level is `AdminLevel.None`, which is the case for any of the DWKit schemes marked as `OwnerAdmin` or `NoAdmin`, the transitions mentioned above that would require an OrgAdmin instead require the project owner.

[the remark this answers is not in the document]
FyreByrd: `RequiredAdminLevel` seems like a reasonable name. I will also clarify the comments with something along the lines of the above comment.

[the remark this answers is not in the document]
7dev7urandom: Ok, I see. My suggestion might be `AdminVerificationLevel` then

FyreByrd:
I have created a wrapper class for `DefaultWorkflow` and have updated all the rest of the code in accordance with that change.
I have also fixed a few small things that I noticed while working on that change (they are related to workflow, so still valid for the PR).

source/SIL.AppBuilder.Portal/common/workflow/index.ts
```diff
@@ -0,0 +1,449 @@
...
+  private productType: ProductType;
+
+  constructor(productId: string) {
+    this.productId = productId;
+    this.currentState = null;
+    this.flow = null;
+  }
+
+  /* PUBLIC METHODS */
+  /** Creat a new workflow instance and populate the database tables. */
```
7dev7urandom, line 43:
```suggestion
  /** Create a new workflow instance and populate the database tables. */
```

source/SIL.AppBuilder.Portal/common/public/workflow.ts
```diff
@@ -0,0 +1,138 @@
...
+ * Examples:
+ *  - If the flow has `WorkflowAdminLevel.High` it will include extra state to represent the organizational approval process
+ *  - If the flow has `WorkflowAdminLevel.Low` it will not include those states, but there are still some states that require action from an OrgAdmin to complete certain actions
+ *  - If the flow has `WorkflowAdminLevel.None` none of the states or actions for the workflow instance will require an OrgAdmin.
+ * 
+ * Any state or transition can have a list of specified `WorkflowAdminLevel`s. What this means is that those states and transitions will be included in a workflow instance ONLY when the instance's `WorkflowAdminLevel` is in the state's or transition's list.
+ * 
+ * If a state or transition does not specify any `WorkflowAdminLevel` it will be included (provided it passes other conditions not dependent on `WorkflowAdminLevel`).
+*/
+export enum WorkflowAdminLevel {
```
7dev7urandom, line 21: I'd still rather this have the word "Verification" or "Required" in it somewhere. As is this makes me think that the workflow has admin privileges, not that it requires admin input.

source/SIL.AppBuilder.Portal/common/workflow/index.ts
7dev7urandom, line 1: Correct me if I'm wrong, but it looks to me like the Workflow class can either be a dummy class that just contains a product id and some helper functions for pulling information out of the db, or it can be a full class with the snapshot information that can be used to actually handle workflows. The former case would be `(new Workflow(productId)).getSnapshot()` which sets the adminLevel and productType but then just returns the snapshot information. The latter would be `(new Workflow(productId)).restore()` which can then be used with `.send()` or similar things.

In my opinion, it's an anti-pattern to have a class that can exist in two possible ways like that. My suggestion would be that the class should only ever exist in the latter form (use an async static function `create` or something to instantiate and populate the class), and that either `.getSnapshot()` becomes a static function with productId as a parameter, or gets taken out entirely such that to get the snapshot information you would instantiate the class as a whole and then take snapshot as a property or similar. 

Does that make sense?

7dev7urandom (changes requested):
Wow, it does look a lot better and easier to use. Let me know what you think of my feedback.

source/SIL.AppBuilder.Portal/common/workflow/index.ts
FyreByrd: I think so. I'll look into it.

FyreByrd: I made the constructor private and converted `create` and `restore` to static functions that would return a Workflow instance (I think this is called the factory pattern?).

source/SIL.AppBuilder.Portal/common/workflow/index.ts
```diff
@@ -0,0 +1,446 @@
...
+  /**
+   * Stops the current running workflow.
+   * 
+   * Note: This does not mean that the workflow is terminated.
+   */
+  public stop(): void {
+    this.flow?.stop();
+  }
+
+  /** Retrieves the workflow's snapshot from the database and sets the `WorkflowAdminLevel` and `ProductType` */
```
7dev7urandom, line 100:
```suggestion
  /** Retrieves the workflow's snapshot from the database and sets the `RequiredAdminLevel` and `ProductType` */
```

7dev7urandom (approved):
Looks good to me, let's try it out!

source/SIL.AppBuilder.Portal/src/routes/(authenticated)/admin/workflows/[product_id]/+page.svelte
```diff
@@ -0,0 +1,230 @@
...
+          Last Transition: {data.instance?.Product.ProductTransitions[0].InitialState}
+          {data.instance?.Product.ProductTransitions[0].Command
+            ? `(${data.instance?.Product.ProductTransitions[0].Command})`
+            : ''}
+        </li>
+        <li>
+          Date: {data.instance?.Product.ProductTransitions[0].DateTransition?.toLocaleTimeString()}
+        </li>
+      </ul>
+      <form
```
Comment (FyreByrd) [inline source/SIL.AppBuilder.Portal/src/routes/(authenticated)/admin/workflows/[product_id]/+page.svelte:107]: Why not superforms?

source/SIL.AppBuilder.Portal/src/routes/(authenticated)/tasks/[product_id]/+page.svelte
```diff
@@ -7,20 +7,21 @@
 </script>
 
 <div class="p-5">
-  <form>
+  <form method="POST">
```
Comment (FyreByrd) [inline source/SIL.AppBuilder.Portal/src/routes/(authenticated)/tasks/[product_id]/+page.svelte:10]: superforms?

FyreByrd (commented):
Switch to using svelte-kit superforms in a few pages

chrisvire (approved):
👍

PR #1021, opened 2024-10-25
Feature/prisma auto date

FyreByrd:
I am tired of having to manually handle the `DateCreated` and `DateUpdated` fields that exist on many of the db models when Prisma is able to handle this through an opt-in feature.
I have updated the prisma schema to opt in to that feature for each field that makes sense to use it on.
I have updated the database writes that can benefit from this in the current codebase, but I have not yet updated the many other writes that would benefit, as they will not be able to benefit until this is merged.


chrisvire (approved):
👍

PR #1025, opened 2024-10-30
Split ProjectSelector functionality into ProjectCard

FyreByrd:
One of the components, `ProjectSelector` was used in two separate locations (projects and project directory) with moderately different use cases. The differences between these use cases led to some of `ProjectSelector`'s implementation being quite janky. It also relied on information returned from the server, which I don't think is a great design choice (unless that data is returned by the root `+layout.server.ts`, as is the case with the user session and the organization list), as there is no clear way to indicate that the component relies on having certain data being returned from the server.

As such, I split the main UI portion into a `ProjectCard` component, which renders the information for a single project, and then duplicated the rest of the `ProjectSelector` code into the two pages the component was originally used in and made the necessary modifications in each location to have only the code that was needed. 

Yes, there is still a decent amount of overlap/duplication; however, that will mostly be resolved in [feature/paginate-projects](https://github.com/sillsdev/appbuilder-portal/tree/feature/paginate-projects) which moves all of the sorting and filtering boilerplate to the server-side.

Yes, I could have just kept this as part of [feature/paginate-projects](https://github.com/sillsdev/appbuilder-portal/tree/feature/paginate-projects), but some of the changes here are significant enough that I anticipated that it would conflict with changes in one of my other branches where I am adding more functionality in the projects page (archival), and I decided to break this out separately so I could rebase that other branch onto this one and resolve those conflicts now, rather than wait for the conflicts to be even more complicated.


source/SIL.AppBuilder.Portal/src/routes/(authenticated)/directory/+page.svelte
```diff
@@ -17,45 +17,64 @@
   let productDefinitionFilter: string;
   let organizationFilter: string;
   let updateDates: [Date, Date] | null;
-</script>
 
-<ProjectSelector
-  projects={data.projects.filter((proj) => {
+  let searchTerm: string = '';
```
7dev7urandom, line 21: As far as I can tell `searchTerm` isn't set anywhere. Is there a place for the user to actually type in a search query somewhere?

source/SIL.AppBuilder.Portal/src/routes/(authenticated)/projects/[filter=projectSelector]/[id]/+page.svelte
```diff
@@ -1,8 +1,96 @@
...
+
+  $: filteredProjects = data.projects.filter((project) => {
+    const searchTermLower = searchTerm.toLowerCase();
+    return (
+      project.Name?.toLowerCase().includes(searchTermLower.toLowerCase()) ||
+      project.Language?.toLowerCase().includes(searchTermLower) ||
+      project.OwnerName?.toLowerCase().includes(searchTermLower) ||
+      project.OrganizationName?.toLowerCase().includes(searchTermLower) ||
+      project.GroupName?.toLowerCase().includes(searchTermLower)
+    );
```
Comment (7dev7urandom) [inline source/SIL.AppBuilder.Portal/src/routes/(authenticated)/projects/[filter=projectSelector]/[id]/+page.svelte:24]:
```suggestion
    return [
      project.Name,
      project.Language,
      project.OwnerName,
      project.OrganizationName,
      project.GroupName
    ].some((field) => field?.toLowerCase().includes(searchTermLower));
```

7dev7urandom (commented):
Looks good to me for the most part. 
I'm curious the reasoning behind moving components into the lib dir. Is there a clear advantage to having them there instead of in the dir they will be used?

source/SIL.AppBuilder.Portal/src/routes/(authenticated)/directory/+page.svelte
FyreByrd: I somehow missed that. Fixed

source/SIL.AppBuilder.Portal/src/routes/(authenticated)/projects/[filter=projectSelector]/[id]/+page.svelte
Comment (FyreByrd) [reply source/SIL.AppBuilder.Portal/src/routes/(authenticated)/projects/[filter=projectSelector]/[id]/+page.svelte:24]: Didn't know you could do it like that. Added to both `/projects` and `/directory`

FyreByrd:
Merging #1036 caused conflicts, so had to rebase and force-push this

7dev7urandom (approved):

FyreByrd:
Fixed the button layout in the UI to be more mobile friendly.
No major changes to how it looks non-mobile, so just including the mobile screenshots.

#### Directory

Before:
image

After:
image

#### Projects

Before:
image

After:
image

chrisvire (approved):
Trusting your judgement

PR #1026, opened 2024-10-30
Add workflow configuration data to WorkflowDefinitions table

FyreByrd:
Added two fields to `WorkflowDefinitions`:
- `ProductType`: an integer that represents the `ProductType` enum in `common/public/workflow.ts`
- `AdminRequirements`: an array of integers that represent the `WorkflowAdminRequirements` enum in `common/public/workflow.ts`

### DIFF
--- a/source/SIL.AppBuilder.Portal/common/prisma/schema.prisma
+++ b/source/SIL.AppBuilder.Portal/common/prisma/schema.prisma
@@ -484,6 +484,8 @@ model WorkflowDefinitions {
   RepublishWorkflows   ProductDefinitions[] @relation("ProductDefinitions_RepublishWorkflowIdToWorkflowDefinitions")
   Workflows            ProductDefinitions[] @relation("ProductDefinitions_WorkflowIdToWorkflowDefinitions")
   StoreType            StoreTypes?          @relation(fields: [StoreTypeId], references: [Id], onDelete: Restrict, onUpdate: NoAction, map: "FK_WorkflowDefinitions_StoreTypes_StoreTypeId")
+  ProductType          Int                  @default(0)
+  AdminRequirements    Int[]                @default([0])
 
   @@index([StoreTypeId], map: "IX_WorkflowDefinitions_StoreTypeId")
 }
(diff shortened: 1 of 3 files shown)


FyreByrd:
I am not 100% sold on the fields being integers. It would be somewhat more readable if they were strings. This would probably require changing the underlying data type of the enums to be strings, which could possibly cause a performance hit for using them as often as they are used.

chrisvire:
@FyreByrd Is this to get these config out of the TypeScript code and into the database?  For future Workflow Definitions, we will need UI to edit/add these settings?

FyreByrd:
> @FyreByrd Is this to get these config out of the TypeScript code and into the database?  For future Workflow Definitions, we will need UI to edit/add these settings?

Yes. I do intend to add UI for that as well, but I wasn't sure whether that should be part of this PR or a different one. I've been trying to keep DB changes as separate minimal PRs.

chrisvire:
@FyreByrd Yes, I think it is good to have separate PRs. You could use interactive rebasing to have the database in one commit and the UI in one comment. It is just a matter of making sure to do a Rebase and merge instead of Squash and merge. Realistically, separate PRs are probably better.

source/SIL.AppBuilder.Portal/common/prisma/schema.prisma
```diff
@@ -484,6 +484,8 @@ model WorkflowDefinitions {
   RepublishWorkflows   ProductDefinitions[] @relation("ProductDefinitions_RepublishWorkflowIdToWorkflowDefinitions")
   Workflows            ProductDefinitions[] @relation("ProductDefinitions_WorkflowIdToWorkflowDefinitions")
   StoreType            StoreTypes?          @relation(fields: [StoreTypeId], references: [Id], onDelete: Restrict, onUpdate: NoAction, map: "FK_WorkflowDefinitions_StoreTypes_StoreTypeId")
+  ProductType          Int                  @default(0)
+  AdminRequirements    Int[]                @default([0])
```
chrisvire, line 488: Do you want the default to be an array containing a value of 0 or an empty array? Is it possible to have an empty array?

FyreByrd: The reason I chose to do an array with a zero rather than an empty array is because zero is the value of one of the enum members. However, given that there should never be a workflow definition where an empty array could be meaningfully distinct from an array with the zero in it, I suppose that that change could be made without too much issue?

chrisvire: What does `[0]` indicate?

FyreByrd: It represents the absence of any specified WorkflowAdminRequirements. This is more important for the state/transition filtering and probably doesn't need to be specified in the WorkflowDefinitions table. So we could change to having an empty array as default and it would work the same if I treat an empty array the same as I would [WorkflowAdminRequirements.None (0)] in the code.

chrisvire: OK. Then it is probably fine.

chrisvire (approved):
👍

FyreByrd:
We may want to remove `WorkflowScheme` and `WorkflowBusinessFlow` from the `WorkflowDefinitions` table in the future. Of course, those will remain for now, in keeping with backwards compatibility with S1 during development.

PR #1027, opened 2024-10-31
Refactor WorkflowInstances to split up stored data

FyreByrd:
This depends on PR #1026 and should not be merged until that PR has been merged and this one has been rebased.

PR #1026 added configuration data for the Workflow state machine directly to the `WorkflowDefinitions` table.
This change means that it is no longer necessary to store that data in `WorkflowInstances` and can instead be accessed through an associated `WorkflowDefinition`.
Pursuant to that, the `WorkflowInstances` table has been modified according to the following:
- `State` column to store the name of the current state of the instance
- `Context` column to store just the necessary parts of the workflow instance's context
- `WorkflowDefinitionId` to reference the associated `WorkflowDefinition`
- `DateCreated` and `DateUpdated`
- deleted `Snapshot` column, since it was split into `State`, `Context`, and the associated fields in `WorkflowDefinitions`

The relationship to `Products` was also changed to be one `Product` to many `WorkflowInstances` in anticipation of the future addition of republish/rebuild workflows, of which there can be many associated with one product.

### DIFF
--- a/source/SIL.AppBuilder.Portal/common/databaseProxy/Products.ts
+++ b/source/SIL.AppBuilder.Portal/common/databaseProxy/Products.ts
@@ -69,7 +69,7 @@ export async function update(
 function deleteProduct(productId: string) {
   // Delete all userTasks for this product, and delete the product
   return prisma.$transaction([
-    prisma.workflowInstances.delete({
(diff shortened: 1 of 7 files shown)


chrisvire (approved):
👍

PR #1028, opened 2024-10-31
Allow hamburger icon in navbar to close sidebar on mobile

FyreByrd:
There was no good way to close the sidebar in mobile without choosing a navigation link in the sidebar, which is less than ideal UX.
I was able to fix this issue by moving the margin-top from the menu links within the sidebar to just the sidebar itself, so it would no longer be layered on top of the navbar and prevent clicks on the navbar from closing the sidebar.


chrisvire:
@FyreByrd With UI changes, could you include before/after screenshots?

7dev7urandom (approved):

chrisvire:
@FyreByrd You don't need to wait for my approval...

PR #1029, opened 2024-11-01
Improved UI for WorkflowDefinition Creation/Editing

FyreByrd:
Updates the UI for `WorkflowDefinitions` creation and editing in line with the database changes in PR #1026.
Also improved the UX for selecting the `WorkflowScheme` and `WorkflowBusinessFlow` fields.

### Changes:

1. Add field for `ProductType`
2. Add field for `WorkflowOptions`
3. Change field for `WorkflowScheme` from text input to dropdown based on `WorkflowScheme` table using the `Code` field.
4. Change field for `WorkflowBusinessFlow` from text input to dropdown
  i.  There wasn't a database table for this one, so I just created a constant array of string values based on what I could find in DWKit directly. I don't anticipate those values changing anytime soon, especially since we should be moving away from DWKit later.

## Comparison Screenshots:

> The UI is practically the same between the creation and editing, and the changes are identical for both, so I am only including screenshots for the creation

### Before

image

### After

image

### After (Changes highlighted)

image


chrisvire:
@7dev7urandom opinions on admin requirements section?

Things that stick out to me:
1. `Administrative Requirement` label style
2. General spacing of checkboxes
3. Mixture of checkboxes and switches

FyreByrd:
> 3. Mixture of checkboxes and switches

I didn't think about that, but yeah, consistency would probably be good. I can change that if need be.

7dev7urandom:
This is an admin only page so I'm not overly concerned about style, but:

- The checkboxes should be toggles (switches). @FyreByrd you might think about giving them a different style than the others, such as a red (`toggle-error`) color which would help show that they are possibly dangerous settings to change. 
- I might suggest a clearer separation of the administrative requirement section (which would slightly change the spacing). Maybe a card essentially where the red circle is in the screenshot would make things look clearer.

FyreByrd:
> This is an admin only page so I'm not overly concerned about style, but:
> 
> * The checkboxes should be toggles (switches). @FyreByrd you might think about giving them a different style than the others, such as a red (`toggle-error`) color which would help show that they are possibly dangerous settings to change.
> * I might suggest a clearer separation of the administrative requirement section (which would slightly change the spacing). Maybe a card essentially where the red circle is in the screenshot would make things look clearer.

How does this look? I used the `warning` semantic color instead of `error`

image

chrisvire:
There are other options to the workflow that is not covered by Administrative Requirements.

* Readiness Check
* Transfer to Authors

Maybe it should just be Workflow Options:
* Show Readiness Check step
* Require organization admin Approval step
* Require organization admin to perform Google Play manual entry steps
* Show Author transitions

FyreByrd:
I added a toggle for the new enum option `AllowTransferToAuthors` and updated the code to be in line with PR #1033 and #1034.

Picture:
image

7dev7urandom:
Looks good from my perspective

FyreByrd:
Made some changes to make the UI more mobile-friendly.

Before Desktop:
image

After Desktop:
image

Before Mobile:
image

After Mobile:
image

chrisvire (approved):
👍

PR #1033, opened 2024-11-05
DB Changes for refactoring workflow state/transition filtration

FyreByrd:
Changes:
- Renamed `AdminRequirements` to `WorkflowOptions` to add more meaningful flexibility in the future for extending the different parameters for the state machine.
- Switched default for `WorkflowOptions` from [WorkflowOptions.None] to [] (Factoring out WorkflowOptions.None completely)

PR #1034, opened 2024-11-06
Refactor filtration of workflow states/transitions

FyreByrd:
Builds on changes made in PR #1033.

Changes:
- Renamed `WorkflowAdminRequirements` to `WorkflowOptions` (future expandability)
- Wrap the filtering object in the `includeWhen` prop of the state/transition meta (readability)
- Refactored the filtration interface to be both more flexible and readable (albeit less performant, which could potentially be improved by updating the backend to a version of NodeJS that supports Set operations).
- Renamed `Workflow.filterMeta` to `includeStateOrTransition` and moved out of `Workflow` class.
- Removed `ReadinessCheck` as a state. (requested)
- Added `AllowTransferToAuthors` as another configuration option in `WorkflowOptions`. (requested)
  - This option is disabled by default across the provided workflow definitions in the seed script, meaning that it has to explicitly be enabled in a workflow definition to allow the `Transfer to Authors` action.

### DIFF
--- a/source/SIL.AppBuilder.Portal/common/public/workflow.ts
+++ b/source/SIL.AppBuilder.Portal/common/public/workflow.ts
@@ -146,7 +146,7 @@ export type WorkflowTransitionMeta = {
  *  - no conditions are specified
  *  - all specified conditions are met
  */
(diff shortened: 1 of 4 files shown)


chrisvire (approved):
👍

PR #1035, opened 2024-11-06
Upgrade project to use Node v22

FyreByrd:
Upgraded the project's node dependency from Node v18/20 to Node v22.


FyreByrd:
The UI does not appear to have any trouble running locally.
Database queries are still working.
BullMQ is also still working.

I still was unable to build, due to the same errors as before regarding the Prisma Client runtime.

PR #1036, opened 2024-11-08
Fix missing i18n

FyreByrd:
Removed `imported-locales` and `flatten.ts`.

Added i18n for:
- Workflow Instances


chrisvire (approved):
👍

source/SIL.AppBuilder.Portal/src/lib/locales/en-us.json
```diff
@@ -428,7 +438,14 @@
   "admin_settings_buildEngines_disconnected": "Disconnected",
   "admin_settings_buildEngines_status": "Status",
   "admin_settings_buildEngines_lastUpdated": "Last update",
-  "admin_workflowInstances_title": "Workflow Instances",
+  "admin_settings_jobAdministration_title":"Job Administration",
+  "admin_settings_jobAdministration_test_instructions":"Choose how long the test task should take",
+  "admin_settings_jobAdministration_test_seconds":"seconds",
+  "admin_settings_jobAdministration_test_add":"Add Test task",
+  "admin_settings_jobAdministration_list":"View Job Statuses",
```
FyreByrd, line 445: Is `Job Administration` etc. what we want to go with for this? it was BullMQ before, but I thought it would be good to be more generic

source/SIL.AppBuilder.Portal/src/lib/timeUtils.ts
```diff
@@ -31,9 +31,9 @@ export function getRelativeTime(date: Date | null) {
   return 'ERROR';
 }
 export function getTimeDateString(date: Date | null) {
-  return `${date?.toLocaleDateString() ?? '-'} ${
+  return `${date?.toLocaleDateString(languageTag()) ?? '-'} ${
```
FyreByrd, line 34: Do we want to use this or no? I have no idea whether this would ultimately be more efficient or not.

source/SIL.AppBuilder.Portal/src/lib/locales/en-us.json
7dev7urandom: Doesn't matter to me. Looks good.

source/SIL.AppBuilder.Portal/src/lib/timeUtils.ts
7dev7urandom: I don't see efficiency being an issue here. I like having `languageTag()` here; very good.

7dev7urandom (approved):
LGTM!

PR #1039, opened 2024-11-13
Pagination UI component

FyreByrd:
Added a UI toolbar for selecting a page out of paginated server data.
Since pagination is not actually implemented in this PR, there wasn't anywhere to show it in use, so I created a temporary page to get some screenshots.
This is its own PR, because it is the base for multiple other future PRs that add pagination in various routes.

#### Desktop

image

#### Mobile

image

### DIFF
--- a/source/SIL.AppBuilder.Portal/src/lib/components/Pagination.svelte
+++ b/source/SIL.AppBuilder.Portal/src/lib/components/Pagination.svelte
@@ -16,11 +16,6 @@
 </script>
 
 <div class="flex flex-row flex-wrap gap-1 w-full max-w-xs md:w-auto md:max-w-none">
-  <span class="input input-bordered flex items-center gap-2 max-w-xs">
-    Show: <!-- TODO: i18n -->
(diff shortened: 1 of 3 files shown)


source/SIL.AppBuilder.Portal/src/lib/components/Pagination.svelte
```diff
@@ -0,0 +1,97 @@
...
+    else return page + i - 1;
+  }
+</script>
+
+<div class="flex flex-row flex-wrap gap-1 w-full max-w-xs md:w-auto md:max-w-none">
+  <span class="input input-bordered flex items-center gap-2 max-w-xs">
+    Show: <!-- TODO: i18n -->
+    <input type="number" name="size" bind:value={size} />
+    / {total}
+  </span>
```
FyreByrd, line 23: I wasn't sure how best to actually do this, so I have refrained from adding anything to the locale JSON until we have discussed this.

FyreByrd (commented):
I left a comment explaining why I did not add a user-facing string to the locale JSON yet

PR #1050, opened 2024-11-20
Use Set methods for state/transition filtration

FyreByrd:
Now that the codebase has been upgraded to depend on Node 22, Set operations are supported server-side.


chrisvire (approved):
👍

PR #1051, opened 2024-11-22
Paginate /product/[id]/files

FyreByrd:
Added server-side pagination and improved the overall UI of `/product/[id]/files`

Notes: 
- The i18n library we have been using currently does not support pluralization, so some of the UI is still a bit rough around the edges.
- All but the latest build are collapsed by default
- The page will show up to 10 builds per pagination by default
- On mobile, the tabular list of the product files is horizontally scrollable, and the name of the file is kept visible

Desktop:

image

Mobile:

image


chrisvire:
@FyreByrd Please resolve merge conflicts.

If we want to add pagination to builds, we can have fewer builds per page (like 2 or 3) since most people only care about the most recent.

FyreByrd:
Resolved merge conflicts

chrisvire:
@FyreByrd We will need to manually format that label until i18n supports pluralization.

chrisvire:
@FyreByrd I notices that when it scrolls, it cuts off the border on the right:

image

chrisvire:
@FyreByrd What problem are you trying to solve with the pagination of product files?  Does it reduce that data transmitted and help performance of the page?

FyreByrd:
> @FyreByrd What problem are you trying to solve with the pagination of product files? Does it reduce that data transmitted and help performance of the page?

It probably doesn't need to be paginated? It won't send a ton of data unless a product has a lot of builds. 

> If we want to add pagination to builds, we can have fewer builds per page (like 2 or 3) since most people only care about the most recent.

You also asked to show only 2 or 3 per page.

FyreByrd:
> @FyreByrd What problem are you trying to solve with the pagination of product files? Does it reduce that data transmitted and help performance of the page?

But yes, that is roughly the goal of pagination. I don't really have enough test data to see what performance gains there would be.

chrisvire:
UX issues:

1. Clicking anywhere in the header should expand/contract the list of files (if we want to be consistent with S1.  Do we need an expanding/contracting list? The original UI was in a popup window and then moved to a route.  I didn't want to spend a lot of time reworking the UI in React since I knew we would be rewriting it.
2. The text color of the selected item is hard to see.

image

3. If we keep the header, the pluralization needs to be addressed.

FyreByrd:
> UX issues:
> 
> 1. Clicking anywhere in the header should expand/contract the list of files (if we want to be consistent with S1.  Do we need an expanding/contracting list? The original UI was in a popup window and then moved to a route.  I didn't want to spend a lot of time reworking the UI in React since I knew we would be rewriting it.
> 2. The text color of the selected item is hard to see.
> 
> 3. If we keep the header, the pluralization needs to be addressed.

I am currently working on these issues. Thank you for bringing 2. to my attention.

chrisvire (approved):
👍

PR #1053, opened 2024-11-25
Revert the relationship between Products and WorkflowInstances to be 1-1

FyreByrd:
After some discussion, it was determined that a Product should only ever have one active WorkflowInstance.

Also updated the database schema to have the correct default for WorkflowOptions.


chrisvire (approved):
👍

PR #1054, opened 2024-12-02
Paginate Workflow Instances display

FyreByrd:
Added general improvements to the UI for displaying workflow instances, as well as adding server-side pagination, search, and sorting.

### Before:

Workflow Instances:
image

Individual:
image

### After:

Workflow Instances:
image

Individual:
image

Details menu:
image


chrisvire (approved):
👍

PR #1059, opened 2024-12-05
Create, Import, and Archive/Reactivate Projects

FyreByrd:
Added jobs to handle creation of projects in BuildEngine.
Handle Project Imports.
I am not 100% certain what the full expected behavior is for Archival and Reactivation, so we will need to discuss this before merging.

Add Project Page:
image

image

Import Project Page:
image

image

### DIFF
--- a/source/SIL.AppBuilder.Portal/common/build-engine-api/requests.ts
+++ b/source/SIL.AppBuilder.Portal/common/build-engine-api/requests.ts
@@ -9,6 +9,31 @@ export async function request(
 ) {
   try {
     const { url, token } = auth.type === 'query' ? await getURLandToken(auth.organizationId) : auth;
+    const check = await prisma.systemStatuses.findFirst({
+      where: {
+        BuildEngineUrl: url,
+        BuildEngineApiAccessToken: token
+      },
+      select: {
+        SystemAvailable: true
+      }
+    });
+    if (!check?.SystemAvailable) {
+      return new Response(
+        JSON.stringify({
+          responseType: 'error',
+          name: '',
+          status: 500,
+          code: 500,
+          message: `System ${url} unavailable`,
+          type: ''
+        } as Types.ErrorResponse),
+        {
+          status: 500,
+          statusText: 'Internal Server Error'
+        }
+      );
+    }
     return await fetch(`${url}/${resource}`, {
       method: method,
       headers: {
--- a/source/SIL.AppBuilder.Portal/common/bullmq/types.ts
+++ b/source/SIL.AppBuilder.Portal/common/bullmq/types.ts
@@ -50,6 +50,7 @@ export enum JobType {
   // Project Tasks
   Project_Create = 'Create Project',
   Project_Check = 'Check Project Creation',
+  Project_ImportProducts = 'Import Products for Project',
   // Publishing Tasks
   Publish_Product = 'Publish Product',
   Publish_Check = 'Check Product Publish',
@@ -103,6 +104,12 @@ export namespace Project {
     workflowProjectId: number;
     organizationId: number;
     projectId: number;
+  }
+
+  export interface ImportProducts {
+    type: JobType.Project_ImportProducts;
+    organizationId: number;
+    importId: number;
   }
 }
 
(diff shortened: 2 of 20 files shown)


chrisvire:
When a project is first created:

image

FyreByrd:
> When a project is first created:
> 
> image

Taken care of

chrisvire:
## Add Project
User Experience issues:
* I accidentally hit enter when the Save button didn't have focus and the project was created. I was surprised by this. This is not the behavior of S1.
* In S1, the save button is disabled until all the required fields (Project Name, Language Code ... and I assume Project Group and Type, but they have default values) are filled in.
* In S2. the only option is "Scripture App Builder" but it should have all the options (in table Application Types)
* In S2, the fields should stack all vertically if on a narrow device

I noticed on the Add Project page that the breakpoint for switching to a narrow layout is 380px. That is not one of the default breakpoints in Tailwind. Curious about it. The breakpoint in S2 is 640 (sm breakpoint).

## Import Project
User Experience issues:
* In S2. the only option is "Scripture App Builder" but it should have all the options (in table Application Types)
* I selected a file and clicked save, but nothing happened.

chrisvire (changes requested):
User experience issues

FyreByrd:
> * In S2. the only option is "Scripture App Builder" but it should have all the options (in table Application Types)

Regarding the available Application Types, I am only presenting the available options based on the `OrganizationProductDefinitions` table. My assumption was that we wouldn't want for users to create a project that isn't valid for their organization. In the example data for dev, all of the entries in `OrganizationProductDefinitions` for SIL have product definitions that use the SAB Application Type, hence SAB is the only option that will be visible.

FyreByrd:
> > * In S2. the only option is "Scripture App Builder" but it should have all the options (in table Application Types)
> 
> Regarding the available Application Types, I am only presenting the available options based on the `OrganizationProductDefinitions` table. My assumption was that we wouldn't want for users to create a project that isn't valid for their organization. In the example data for dev, all of the entries in `OrganizationProductDefinitions` for SIL have product definitions that use the SAB Application Type, hence SAB is the only option that will be visible.

This is incorrect behavior and has been updated to show all application types

FyreByrd:
> * I accidentally hit enter when the Save button didn't have focus and the project was created. I was surprised by this. This is not the behavior of S1.

I have resolved all the issues except this one. As best as I can tell, this is the default behavior of the browser. This would likely need to be fixed in multiple location, not just the add project form.
For now, I think the issue has been partially mitigated by blocking form submission unless the name and language code are both set.

chrisvire:
> > * I accidentally hit enter when the Save button didn't have focus and the project was created. I was surprised by this. This is not the behavior of S1.
> 
> I have resolved all the issues except this one. As best as I can tell, this is the default behavior of the browser. This would likely need to be fixed in multiple location, not just the add project form. For now, I think the issue has been partially mitigated by blocking form submission unless the name and language code are both set.

I think my major concern was that the save button was enabled when the required fields were not all filled out.

FyreByrd:
> > > * I accidentally hit enter when the Save button didn't have focus and the project was created. I was surprised by this. This is not the behavior of S1.
> > 
> > 
> > I have resolved all the issues except this one. As best as I can tell, this is the default behavior of the browser. This would likely need to be fixed in multiple location, not just the add project form. For now, I think the issue has been partially mitigated by blocking form submission unless the name and language code are both set.
> 
> I think my major concern was that the save button was enabled when the required fields were not all filled out.

In that case the issue should be fixed by https://github.com/sillsdev/appbuilder-portal/pull/1059/commits/2021603abb81dd3f28074fc00d1a3cfb9a955394

chrisvire (changes requested):
## Add Project
User Experience issues:
1. The Add Products button should be disabled until the repository location is available.

## Import Project
User Experience issues:
1. I selected a group, type, and chose a file, but the save button was disabled so I couldn't do anything.

image

FyreByrd:
> ## Import Project
> User Experience issues:
> 
> 1. I selected a group, type, and chose a file, but the save button was disabled so I couldn't do anything.
> 
> image

I was unable to reproduce this issue. I am fairly certain that it should be showing some error messages if it is in a state that would block submission.

FyreByrd:
Added kebab menus to both the projects list and the individual project view:

image

image

PR #1062, opened 2025-01-15
Fix color scheme in CSS themes

FyreByrd:
Per @chrisvire 's request, I am opening this PR to fix some of the issues in the color scheme with the light vs. dark theme.
While I only have one fix right now, I think we should discuss some other fixes to the color scheme.

### DIFF
--- a/source/SIL.AppBuilder.Portal/src/lib/components/settings/MultiselectBoxElement.svelte
+++ b/source/SIL.AppBuilder.Portal/src/lib/components/settings/MultiselectBoxElement.svelte
@@ -7,7 +7,7 @@
 <div class="my-2">
   <label>
     <span class="flex items-center">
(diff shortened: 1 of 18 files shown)


FyreByrd:
We had been using `base-300` for some of the backgrounds, because `neutral` was too dark in light mode.
Per @chrisvire 's request, I significantly lightened the `neutral` color in light mode.
I then went ahead and applied it in several locations where `base-300` had previously been applied.

Before:
image

After:
image

FyreByrd:
We should probably discuss what we want to do with the `accent` color. 

It is currently a mauve-ish color `#e0add2` in light and `#987c90` in dark.

We don't currently use it anywhere. We may want to consider changing the color and then I might use it in some places where I had been using `primary`.

FyreByrd:
New Accent color:

Light:
image

Dark:
image

chrisvire:
@FyreByrd Try neutral as `#f3f3f3` which is used in S1

PR #1065, opened 2025-01-30
Sort named items in admin

chrisvire:


FyreByrd (approved):
It looks good to me. There are probably a few more places that will need this treatment, but server-side pagination in the places I can think of will probably require a different approach and would be out of scope for this PR.

PR #1067, opened 2025-02-03
Rebuild/Republish products from UI 

FyreByrd:
- Expanded `StartupWorkflow` to also include rebuild and republish
- Fixed a few issues stemming from deleting workflow instances
- Added UI for rebuilding/republishing products

Notes:
- This does NOT implement automatic rebuild (we can talk about adding that to this PR)
- I currently have it so that bulk rebuild will ignore products in the backend that can't be rebuilt. Not sure if this is the desired behavior. However, since the frontend has the same constraints, it shouldn't be an issue.

Bulk Modal:

image

Individual product menu:

image


FyreByrd:
I did make some changes from how the bulk rebuild modal is handled in S1.
The modal in S1 simply displays a list of all product definitions that are in the selected projects with no breakdown by project, no indication if a product definition has either rebuild or republish available, and no regard for if the product can be rebuilt.

S1 with no products selected:
image

S1 with an invalid selection:
image

In contrast, the implementation I currently have for S2 groups the product definitions by project, shows whether a product has a rebuild or republish workflow defined, and only displays products that are actually able to be rebuilt/republished.

S2 with no products selected:
image

S2 with an invalid selection: not possible

FyreByrd:
Rebased on top of feature/svelte. Squashed a few commits for future readability

chrisvire (approved):
👍

PR #1073, opened 2025-02-26
Individual project view subroute in /directory

### LINKED ISSUE #1071: Individual project display in directory
S2 currently only has a list of projects in the directory route.

In S1, the links for each project in the list go to a new page showing just that project.

The individual page includes info about the organization, and then the build info (i.e. artifacts) for each product in the project.

Image

FyreByrd:
Fixes #1071 

Screenshot from 2025-02-26 13-05-17

Currently only shows the latest build associated with a successful release. We may want to change this to be more in line with S1, but we would need to discuss implementation details and use cases. I don't exactly like the idea of returning a ton of data (i.e. all artifacts for all builds for all products for a project) when only some of it will be useful to the end user.

One other feature I added is a link to the /projects route for the associated project if the user has the correct permissions.


chrisvire (approved):
👍

PR #1077, opened 2025-03-05
Add artifact filters from DWKit

FyreByrd:
Specific states in the workflow only need to show some of the artifact types. S1 had a way to do that, this PR does the same thing now.

### DIFF
--- a/source/SIL.AppBuilder.Portal/common/workflow/state-machine.ts
+++ b/source/SIL.AppBuilder.Portal/common/workflow/state-machine.ts
@@ -50,8 +50,8 @@ export const WorkflowStateMachine = setup({
     includeFields: [],
     /** Reset to false on exit */
     includeReviewers: false,
-    /** Reset to false on exit */
-    includeArtifacts: false,
+    /** Reset to null on exit */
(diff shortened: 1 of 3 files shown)


FyreByrd:
I couldn't find where in DWKit `ProductLatestApk` was used. My assumption is that it was superseded by `ProductLatestAab`, but I still included it.

source/SIL.AppBuilder.Portal/common/workflow/state-machine.ts
```diff
@@ -720,11 +716,18 @@ export const WorkflowStateMachine = setup({
           }
```
chrisvire, line 716: For the includeFields, you probably only need the listingLanguageCode for Android_GooglePlay.

chrisvire:
You can search the forms at `source/OptimaJet.DWKit.StarterApplication/Metadata`.  

It looks like `ProductLatestApk` is not used anymore. This is due to the fact that the filter is used during "Create App Store Entry" and you can only create new apps using AAB. Please remove `latestAPK`.

chrisvire (approved):
👍

PR #1078, opened 2025-03-05
Add UserId field to ProductTransitions

FyreByrd:
This is something we discussed regarding transitioning away from DWKit.
The WorkflowUserId field is a string that holds a guid associated with the user profile in DWKit. As it currently is, that same field will also be on the associated User. Since we are moving away from DWKit, those fields should eventually be removed. This removal would also remove all links between transitions and the user responsible for the transitions, hence, this PR to add a UserId field that would reference Users.


chrisvire (approved):
👍

PR #1080, opened 2025-03-07
Upgrade from svelte-flatpickr to svelte-flatpickr-plus

FyreByrd:
This fixes the broken dependencies from svelte-flatpickr (which also has not been updated to work with Svelte 5 very well).


FyreByrd:
There is another potential fix on https://github.com/sillsdev/appbuilder-portal/tree/fix/use-pikaday, but I like this fix better. I didn't have to do much extra work for it to fit the only use case we need it for.

PR #1081, opened 2025-03-07
Add wrapper functions for checking user roles

FyreByrd:
I was tired of trying to remember which array index on session was the user role and which was the organization. These functions simplify all of that.

Functions added:
- `isAdminForOrg`: returns true if user is a SuperAdmin, or is an OrgAdmin for the specified organization
- `isAdminForOrgs`: returns true if user is a SuperAdmin, or is an OrgAdmin for any of the provided orgs
- `isAdmin`: returns true if user is a SuperAdmin, or is an OrgAdmin for any organization
- `isSuperAdmin`: returns true if user is a SuperAdmin
- `hasRoleForOrg`: returns true if user has specified role in specified org
- `orgsForRole`: returns a list of organizations where the user has the specified role


chrisvire (approved):
👍

PR #1082, opened 2025-03-07
Remove deprecated features from Svelte 5

FyreByrd:
Svelte 5 deprecated some features that we were using but otherwise still work:

`createEventDispatcher` is superseded by event props

`$app/stores` is superseded by `$app/state`

### DIFF
--- a/source/SIL.AppBuilder.Portal/src/lib/components/Dropdown.svelte
+++ b/source/SIL.AppBuilder.Portal/src/lib/components/Dropdown.svelte
@@ -2,26 +2,32 @@
 @component
 A simple dropdown menu from DaisyUI.
 -->
+<!-- TODO: this component isn't used anywhere... -->
 <script lang="ts">
-  import { createEventDispatcher } from 'svelte';
-  export let cols = 6;
-  const dispatch = createEventDispatcher();
+  interface Props {
+    cols?: number;
+    label?: import('svelte').Snippet;
+    content?: import('svelte').Snippet;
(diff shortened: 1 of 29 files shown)


source/SIL.AppBuilder.Portal/src/lib/components/Dropdown.svelte
```diff
@@ -2,26 +2,32 @@
...
 -->
+<!-- TODO: this component isn't used anywhere... -->
 <script lang="ts">
-  import { createEventDispatcher } from 'svelte';
-  export let cols = 6;
-  const dispatch = createEventDispatcher();
+  interface Props {
+    cols?: number;
+    label?: import('svelte').Snippet;
+    content?: import('svelte').Snippet;
```
chrisvire, line 10: I think you should do:

`import { type Snippet } from 'svelte';`

and then use:

  interface Props {
    cols?: number;
    label?: Snippet;
    content?: Snippet;
    onNavEnd?: () => void;
  }

source/SIL.AppBuilder.Portal/src/routes/(authenticated)/+layout.svelte
```diff
@@ -22,7 +22,7 @@
   }
 
   let orgMenuOpen = false;
-  // $: console.log($page.data);
+  // $: console.log(page.data);
```
chrisvire, line 25: This can be removed

chrisvire (approved):
👍

PR #1083, opened 2025-03-07
Fix bug in organization selection on `/users` and `/workflow-instances`

FyreByrd:
Selecting an org on the dropdown select menu didn't do anything


chrisvire (approved):
👍

PR #1085, opened 2025-03-07
Rename projects common

FyreByrd:
Renamed:
- `$lib/projects/common.ts` => `$lib/projects/index.ts` 
- `$lib/projects/comon.server.ts` => `$lib/projects/server.ts`

Why? I like how the imports look better.

- `import x from '$lib/projects/common` => `import x from '$lib/projects'`
- `import x from '$lib/projects/common.server` => `import x from '$lib/projects/server'`


chrisvire (changes requested):
Please do not use relative paths for imports but use '$lib/...'

chrisvire (approved):
👍

PR #1086, opened 2025-03-07
Remove unneeded try-catch in form submissions

FyreByrd:
Several of the server side form submission handlers were wrapped in a try-catch that was looking for `ValiError`s. None of the functions in any of those handlers will ever throw a `ValiError`.

I also shortened the variable name `superFormData` to `form` in several locations. (this is an artifact from when we used the `form` variable from sveltekit to trigger a redirect. I changed that in Svelte 5 to instead be triggered in the client-side form handler method).


chrisvire (approved):
👍

PR #1087, opened 2025-03-07
Replace { [key: type]: type} with Record<type, type>

FyreByrd:
This is a stylistic thing more than anything. It does not affect the functionality of the code at all. I just think it looks nicer.


chrisvire (approved):
Looks fine.

PR #1088, opened 2025-03-07
Replace reduce with more exact methods

FyreByrd:
I did not know earlier about the JS array methods `every`, `flatMap`, etc.

This improves performance by cutting out an unnecessary extra array looping function call in several locations


chrisvire (approved):

FyreByrd:
Hooray for merge conflicts...

PR #1093, opened 2025-03-18
Fix UseDefaultBuildEngine settings

FyreByrd:
- Default value of `Organization.UseDefaultBuildEngine` should be `true`
- Hide url/token settings if org uses default build engine
- Also improved filtration of url/token pairs from organizations

### DIFF
--- a/source/SIL.AppBuilder.Portal/common/build-engine-api/requests.ts
+++ b/source/SIL.AppBuilder.Portal/common/build-engine-api/requests.ts
@@ -19,20 +19,18 @@ export async function request(
       }
     });
     if (!check?.SystemAvailable && resource !== 'system/check') {
-      return new Response(
-        JSON.stringify({
+      return {
+        ok: false,
+        status: 500,
+        json: {
           responseType: 'error',
           name: '',
           status: 500,
           code: 500,
           message: `System ${url} unavailable`,
           type: ''
-        } as Types.ErrorResponse),
-        {
-          status: 500,
-          statusText: 'Internal Server Error'
-        }
-      );
+        } as Types.ErrorResponse
+      };
     }
     return await fetch(`${url}/${resource}`, {
       method: method,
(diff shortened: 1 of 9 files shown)


chrisvire:
It is weird that UseDefaultBuildEngine is nullable. That is how it is. So we have to deal with it.

chrisvire:
@FyreByrd I notice a difference in S1 and S2 with the SystemStatuses table.  With S1, the DateUpdated is changed when the SystemAvailable value changes.

So in app.scriptoria.io/admin/settings/build-engines, the "Last update" was when there was a change in value so not the last time it was checked but more of an uptime value.

image

chrisvire:
Maybe there would be value if having a "Last checked" vs "Uptime", but that would require another field in the table.

FyreByrd:
> Maybe there would be value if having a "Last checked" vs "Uptime", but that would require another field in the table.

I like the idea of that. Perhaps a future PR?

FyreByrd:
> With S1, the DateUpdated is changed when the SystemAvailable value changes.

Updated accordingly

chrisvire:
It is not detecting and reporting when it is down.

chrisvire (changes requested):
How does the system status checking report when it is down and not responding?

Also, in production, the server is sitting behind CloudFlare, so you might get a replacement response when CloudFlare notices that page is not available. We can test that out.

chrisvire:
@FyreByrd I tried your new code and it still leaves 3 entries in SystemStatuses.

FyreByrd:
> Issue:
> 
> 1. when I run bootstrap, it puts "replace" in the BuildEngineApiAccessToken.
> 2. in SystemStatuses, I see 3 entries for these with value SystemAvailable=false
> 3. I updated the token in the organizations and then promoted the check status
> 4. now I see 6 entries (3 each of token=replace, SystemAvailable=false and token=, SystemAvailable=true)
> 
> Expectation: there should only be 1 entry per Url/Token pair there should only be current entries (not past ones)

This should all be fixed now.
I am not certain what effect running S1 on the same db at the same time will have.

Sometimes you will need to run `npm run dev` twice in order to fully refresh the BullMQ code. Not sure why. It may have something to do with how the background process works?

chrisvire (approved):
👍

PR #1094, opened 2025-03-24
Upgrade to Paraglide 2.0

FyreByrd:
Paraglide 2.0 adds several useful features, including pluralization.


FyreByrd:
Theoretically, Paraglide 2 also supports a nested message format, similar to what we had previously. I can reformat our message JSONs to take advantage of that if desired.

FyreByrd:
There is a small issue I noticed with pluralization. It still works; it's just that `npm run check` complains. I plan on opening an issue with Paraglide tomorrow.

FyreByrd:
> There is a small issue I noticed with pluralization. It still works; it's just that `npm run check` complains. I plan on opening an issue with Paraglide tomorrow.

I fixed the problem without having to open an issue. It works without complaint now

source/SIL.AppBuilder.Portal/vite.config.ts
```diff
@@ -56,8 +56,15 @@ export default defineConfig({
         }
       }
     },
-    paraglide({ project: './project.inlang', outdir: './src/lib/paraglide' }),
     sveltekit(),
+    paraglideVitePlugin({
+      project: './project.inlang',
+      outdir: './src/lib/paraglide',
+      // TODO: not sure how we want to specify the strategy
```
FyreByrd, line 63: As far as I can tell, the 'url' strategy matches what we were doing previously. This TODO can probably be removed

source/SIL.AppBuilder.Portal/src/app.html
```diff
@@ -1,5 +1,6 @@
 <!DOCTYPE html>
-<html lang="%paraglide.lang%" dir="%paraglide.textDirection%">
+<!-- TODO: Figure out how to replace textDirection -->
+<html lang="%lang%" dir="%paraglide.textDirection%">
```
FyreByrd, line 3: The variable %paraglide.textDirection% is leftover from the original Paraglide SvelteKit adapter. Paraglide 2 bundles everything into a Vite adapter, but makes no mention anywhere of what to replace textDirection with.

FyreByrd:
> Theoretically, Paraglide 2 also supports a nested message format, similar to what we had previously. I can reformat our message JSONs to take advantage of that if desired.

I tried this. It isn't as nice as I thought it would be. Literally the only advantage is that the JSON file for the messages can be collapsed a bit. It also appears to break the Sherlock VS Code extension, which is very helpful to have.

source/SIL.AppBuilder.Portal/src/app.html
chrisvire: We currently don't have RTL localizations. It is probably fine to remove this.

source/SIL.AppBuilder.Portal/project.inlang/settings.json
```diff
@@ -1,15 +1,16 @@
...
-    "https://cdn.jsdelivr.net/npm/@inlang/message-lint-rule-missing-translation@latest/dist/index.js",
-    "https://cdn.jsdelivr.net/npm/@inlang/message-lint-rule-without-source@latest/dist/index.js",
-    "https://cdn.jsdelivr.net/npm/@inlang/plugin-m-function-matcher@latest/dist/index.js",
-    "https://cdn.jsdelivr.net/npm/@inlang/plugin-i18next@latest/dist/index.js"
+    "https://cdn.jsdelivr.net/npm/@inlang/plugin-message-format@4/dist/index.js",
+    "https://cdn.jsdelivr.net/npm/@inlang/plugin-m-function-matcher@2/dist/index.js"
   ],
-  "plugin.inlang.i18next": {
-    "pathPattern": "./src/lib/locales/{languageTag}.json"
+  "plugin.inlang.messageFormat": {
```
chrisvire, line 13: What is this?

FyreByrd: This is changing the paraglide configuration to the recommended

FyreByrd: Correction, this is what I got after running `npx @inlang/paraglide-js init` (I had to manually change a few things). So I guess it's roughly the default configuration.

7dev7urandom:
Looks good. This makes a lot of things much nicer. 
Not a huge fan of the variant syntax for pluralization but oh well. 
If the url strategy alone works fine then I'm good with this.

7dev7urandom (approved):

chrisvire (approved):
👍

FyreByrd:
We can always look into the nested keys later. That will be a more involved set of changes anyways.

PR #1108, opened 2025-04-02
Various i18n fixes

FyreByrd:
Most of this was adding in some missing translations.

I also altered some more of our translations to use variants, specifically transitionTypes, user_roles, workflowTypes, productTypes, and workflowDefinitions_options, because these key off of enums that we defined in the code and doing it like this makes it really easy to just pass in the enum value as an argument to the translation function.

I forgot to finish up the LanguageCodeTypeahead to use the pluralized variants in #1079 so I fixed that here.


FyreByrd:
I would also be interested in going ahead and shortening the translation keys (and also removing some duplicate values as well) since I doubt we will be going back to nested keys (at least I don't want to go back to those).

chrisvire (approved):
👍

PR #1110, opened 2025-04-04
UI for Publishing Properties

### LINKED ISSUE #1106: Add UI for Product Publishing Properties
S1 Screenshot

Image

FyreByrd:
Fixes #1106 

image

I created a separate component for the textarea specifically so I could also reuse it in the workflow definitions and product definitions UI.

### DIFF
--- a/source/SIL.AppBuilder.Portal/src/lib/components/settings/PropertiesEditor.svelte
+++ b/source/SIL.AppBuilder.Portal/src/lib/components/settings/PropertiesEditor.svelte
@@ -11,14 +11,17 @@
 
   let { value = $bindable(), name, className = '', ok = $bindable(true) }: Props = $props();
 
-  let parsed = $derived(safeParse(propertiesSchema, value));
+  const parsed = $derived(safeParse(propertiesSchema, value));
 
   let rawValue = $state(value);
 
   $effect(() => {
-    if (parsed.success) {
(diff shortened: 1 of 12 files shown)


chrisvire:
I tried clicking "Default", causing an error in the JSON, and then tried to save it.

image

I clicked cancel and then opened the dialog again to try something else and it still has the same data:

image

I tried fixing it, but it still shows the error and the save button is still disabled.

image

It didn't clear the error or re-enable the save button until blur (I had to click outside the text area).

Please fix:
1) reset the text area (or the whole dialog) on cancel 
2) clear the error and re-enable the save button when the text is changed (not waiting for blur).

chrisvire (changes requested):
Please fix:

* reset the text area (or the whole dialog) on cancel
* clear the error and re-enable the save button when the text is changed (not waiting for blur).

chrisvire:
Issue: compute type dropdown is not reset after cancel.
Steps to reproduce:
1. select Publishing Properties
2. select medium
3. click cancel
4. select Publishing Properties

The compute type is set to medium and I cannot select medium again.

image

FyreByrd:
I noticed something similar but was unsuccessful in fixing it. I will make another attempt

chrisvire (approved):
👍

PR #1118, opened 2025-04-21
Fix LanguageCodeTypeahead results to more closely match S1

### LINKED ISSUE #1099: Differences in lanaguage selector
With the new language selector, there are differences in how it is implemented.

Here are a few things I noticed:

1) Searching for alternate names doesn't find the correct item.

Here is an example of searching for "Alevica"

S1

Image

S2

Image

2) Displaying an entry is different

Here is an example of searching for "Zazaki"

S1

Image

S2

Image

Do we wait for the completion of #1076 ? Can we release with this until that issue is completed?

Here is what these searches look like in the React version of the new component:

Image

Image

FyreByrd:
Fixes #1099 

Granted, it still isn't identical to S1, but it is much closer.

S1:
image

S2:
image

S1:
image

S2:
image

S1:
image

S2:
image

S1:
image

S2:
image

### DIFF
--- a/source/SIL.AppBuilder.Portal/src/lib/components/LanguageCodeTypeahead.svelte
+++ b/source/SIL.AppBuilder.Portal/src/lib/components/LanguageCodeTypeahead.svelte
@@ -49,7 +49,7 @@
   });
 
   function search(searchValue: string) {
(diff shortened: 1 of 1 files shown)


chrisvire (approved):
Much better!

PR #1119, opened 2025-04-22
Use specific ports instead of defaults

FyreByrd:
Use localhost:6173 for UI
Use localhost:6100 for backend

Since 5173 is the default for SvelteKit projects, and we now have several of those, using different ports for each makes it simpler to have multiple running locally if needed.


chrisvire (approved):
👍

PR #1121, opened 2025-04-23
Hide Download link in Project Directory if downloads not allowed

FyreByrd:
As it turns out, the project directory page was already filtered by if projects were public. This ended up just needing to fix usage of `Project.AllowDownloads`.

Also resolved a TODO comment regarding `Organization.PublicByDefault`.


chrisvire (changes requested):
I see that the download links are hidden, but the information about the URLs to the download files are still sent to the page. I was able to look in the network trace in DevTools and find the URLs (in my case, I was looking at `http://localhost:6173/directory/10` and the data was in the first request/response (i.e. 10 in the network list).

FyreByrd:
The download links are no longer sent to the front end when allow downloads is false. I also made sure to not send any of the information associated with the latest publication.

FyreByrd:
This shouldn't need any further review, as the one outstanding issue with this has been fixed

PR #1122, opened 2025-04-23
Use Prisma to compose typing for components

FyreByrd:
I got tired of having to specify the exact type of specific fields from the database. Prisma.<model>GetPayload creates an easy way to "select" database fields to be included in a type. This was already used in a few other places already.

Also updated the roles utility functions to key off of the Session type from `src/auth.ts` (no substantive change).

If at all possible, I would like this to be merged *before* #1120 so I can make some changes there to reflect this PR.


chrisvire (approved):
👍

PR #1143, opened 2025-05-21
Add OpenTelemetry

FyreByrd:
Sends some trace data to an opentelemeter-collector docker image. Currently the image is configured to just log what is sent. 
Nothing is sent to Honeycomb yet. We will need to implement a sampling strategy before that happens.

I Currently have traces configured for:
- node auto instrumentation (provided by OpenTelemetry)
- hooks.server.ts/auth.ts
- BuildEngine requests
- redis/db connections
- all of the BullMQ workers (but nothing more granular on the individual jobs yet)
- /projects/[id]: load
- /users: load

According to my calculations, with the free tier of Honeycomb, we can only afford to send **7** events per **second** (Honeycomb counts each individual **span** as an event). What I have so far would be sending astronomically more than that, even with only one user testing locally. We need to have a serious discussion about what we are actually interested in sending. We may want to consider a more self-hosted solution for some of the metrics we are interested in.

Sidenote: I have occasionally been using this as a way to better visualize the trace data locally in lieu of the otel-collector image: https://github.com/CtrlSpice/otel-desktop-viewer


7dev7urandom:
Closing in favor of #1186

PR #1147, opened 2025-05-27
Add public user profile page

### LINKED ISSUE #1142: Implement user profile display
S1 has a page, /users/id, that will display some general information about the user. This is missing in S2.

Image

FyreByrd:
Fixes #1142 
I am not sure what the purpose of this page really is. In S1 it requires authentication to view, so it is not truly "public". There are also no internal links in S1 that link to this page as best as I can tell.

I did change a few things from the S1 implementation:
1. Slightly different layout for mobile
2. Use SHA-256 instead of md5 for the gravatar hash (gravatar documentation says to use SHA-256)
   a. Note: this difference is why the avatar images are different between S1 and S2
3. Add edit link with pencil icon if the current user is the subject, or is an admin for the subject.

### Screenshots

S1:
image

S2:
image

S1 mobile:
image

S2 mobile:
image


FyreByrd:
Associated issue was closed as not planned

PR #1151, opened 2025-06-02
Fix transition for successful builds

FyreByrd:
Per online discussion, cleaned up the transition guards for successful builds.

As best as I can understand, we only want to transition to 'Create App Store Entry' iff
1. The product type is Google Play AND
2. The workflow type is startup AND
3. The project has NOT been uploaded to Google Play

In all other cases, the state machine should transition to 'Verify and Publish'


chrisvire (approved):
This works for the rebuild workflow.  We many need to revisit the logic of the state machines to verify the different workflows work with this.

PR #1182, opened 2025-07-28
CI tests on GH actions

7dev7urandom:
ESLint and fixed linting errors
Fix env vars and remove unnecessary VITE prefix
Ignore startup connections such as db/valkey while building


7dev7urandom:
@coderabbitai review

PR #1214, opened 2025-08-30
Update workflow definitions in S1->S2 migration job

### LINKED ISSUE #1211: App Builder Configuration for non-Google Play Android app: Should not have "New App" and "Exisiting App"
On the App Builder Configuration step, the  "New App" and "Existing App" should only be available for apps being published to Google Play.

On staging (https://dev.scriptoria.io/projects/147, Android App to Amazon S3 Bucket), I am getting this:

Image

Here is a similar page for a new Android App to Amazon S3 Bucket on production:

Image

I also noticed in production that the "Store Listing Language" is listed. I think this is OK since the project data has not been uploaded yet.

7dev7urandom:
Closes #1211


FyreByrd (approved):
Looks good to me

PR #1237, opened 2025-09-04
Remove AllowTransferToAuthors

FyreByrd:
Per discussion with @chrisvire removed this as a WorkflowOption. The new code hides author steps only when there are no authors present, rather than also checking config options.


chrisvire (approved):
👍

PR #1305, opened 2025-09-30
Upgrade packages

7dev7urandom:
Upgrade and test all npm packages as well as docker image. Remove legacy express dependency. 
All should be working as far as I can tell, some minor issues may surface as we use it. 

Closes #1190, #1191, #1197, #1199, #1200, #1201, #1203, #1204, #1216, #1229, #1231, #1271, #1273


src/lib/components/OrganizationDropdown.svelte
```diff
@@ -7,19 +7,19 @@
     value: number | null;
     className?: string;
     allowNull?: boolean;
-    [key: string]: unknown;
+    selectProperties?: Record<string, unknown>;
```
FyreByrd, line 10: This could probably be typed as HTMLInputAttributes?

PR #1325, opened 2025-10-02
Fix Organization Selector

### LINKED ISSUE #1320: Change Org Selector UX
We tried to move away from having the organization display at the top left.  What we were really trying to move away from was having the current organization controlled through headers.

While trying to actually use S2, I feel like we need to maintain the organization selector/display in the top left (with  All Organizations).

FyreByrd:
Fixes #1320 

Screenshot 2025-10-02 at 4 42 11 PM

Screenshot 2025-10-02 at 4 42 28 PM

## Summary by CodeRabbit

- New Features
  - Header/drawer now shows a sortable, clickable organization picker; auto-selects when only one org.
  - Organizations page displays an inline table with logos, names and contact emails; rows navigate to settings.

- Bug Fixes / Behavior Changes
  - Projects filter route no longer performs automatic per-request org redirects.
  - Projects, Users and related pages no longer show organization dropdowns; they respect the active organization.

- Refactor
  - Consolidated organization selection into layout and pages; removed intermediate selector components/pages.


chrisvire:
The height of the Organization selector should be the same height as the bar at the top of the app.


image

chrisvire (approved):
👍

PR #1338, opened 2025-10-06
Add step to manually handle publish error for GooglePlay

### LINKED ISSUE #1326: Handle manual upload after addressing app content issue
If there is a policy violation in the app content, the publish will fail.  The user can address the policy violation and manually upload the APK/AAB that failed with the publish.  The user needs to be able to update Scriptoria to say that the publish succeeeded manually.

FyreByrd:
Fixes #1326

Adds a step after Product Publish that is included if the product type is Android GooglePlay. The step is reached if a publish fails and the console text matches `/Google Api Error/i`.

Prevents the happy path action from being submitted on the tasks page if there are any checkboxes in the instructions that are not checked.

## Summary by CodeRabbit

* **New Features**
  * Google Play error-evaluation step with an instructions screen and confirmation checkbox.
  * ReleaseInfo panel showing publish results: channel, status, date, resolved date, and console link.

* **Improvements**
  * Console output inspected to detect Google API errors and route workflows accordingly.
  * Workflow updated to record a "Date Resolved" when issues are marked resolved.
  * Artifact file-size now uses an explicit integer type and is null/empty when unknown.

* **Localization**
  * Added labels for publish results and resolved date.

### DIFF
--- a/src/routes/(authenticated)/tasks/[product_id]/instructions/Evaluate_Error.svelte
+++ b/src/routes/(authenticated)/tasks/[product_id]/instructions/Evaluate_Error.svelte
@@ -1,6 +1,50 @@
-<span class="flex flex-row">
-  <input type="checkbox" class="checkbox checkbox-accent" />
-  &nbsp; I confirm that I have taken the necessary steps to manually resolve the error in the&nbsp;
-  <a href="https://play.google.com/apps/publish/" target="_blank">Google Play Developer Console</a>
-  .
-</span>
+<h3>Instructions</h3>
+<ul>
+  <li>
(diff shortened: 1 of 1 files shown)


chrisvire:
Based on our discussion, we would like to have a field in the workflow UI that requires the user to acknowledge a statement before the continue button is enabled.  To be more generic, if there are check boxes in the instructions, the happy path action is disabled until all the check boxes are checked.

For the Handle Publish Error state, it will update the ProductPublication.DateResolved (new field) when the state is completed. This needs to be displayed in the publication area in the Product Files list.

Issue #1246, opened 2025-09-08
Rebuilds: Enable rebuild on upload


chrisvire:
1) In the route `GET /api/projects/[id=idNumber]`, add a `can-rebuild` attribute to the project result.  This value will be true if:
    * The user requesting the token is the owner of the project 
    * The state of one of the products has completed the initial workflow (i.e. non-null PublishDate and PublishLink) and there isn't a workflow in progress 

2) Provide a route `POST /api/projects/[id=idNumber]/rebuild` that will initiate a rebuild on the project if the above conditions are true.

PR #1342, opened 2025-10-08
Feature/enable rebuild on upload

becca-perk:
Added 'can-rebuild' attribute to the token route and created a new rebuild route to initiate rebuilds. 
Addresses https://github.com/sillsdev/appbuilder-portal/issues/1246


src/routes/(unauthenticated)/api/projects/[id=idNumber]/rebuild/+server.ts
```diff
@@ -0,0 +1,63 @@
+import { json } from '@sveltejs/kit';
+import { ProductActionType } from '$lib/products/index';
+import { doProductAction } from '$lib/products/server';
+import { DatabaseReads } from '$lib/server/database';
+
+/** Wrapper function to return error messages for AppBuilders */
+function createAppBuildersError(status: number, title: string) {
+  return new Response(JSON.stringify({ errors: [{ title }] }), { status });
+}
```
Comment (FyreByrd) [inline src/routes/(unauthenticated)/api/projects/[id=idNumber]/rebuild/+server.ts]: This could probably be extracted to an external common ts file, but it's such a short function that that probably wouldn't be necessary.

src/routes/(unauthenticated)/api/projects/[id=idNumber]/rebuild/+server.ts
```diff
@@ -0,0 +1,63 @@
...
+      DatePublished: true,
+      PublishLink: true,
+      WorkflowInstance: { select: { Id: true } }
+    }
+  });
+
+  // Find rebuildable products: published & no workflow in progress
+  const rebuildableProducts = products.filter(
+    (p) => p.DatePublished && p.PublishLink && !p.WorkflowInstance
+  );
```
Comment (FyreByrd) [inline src/routes/(unauthenticated)/api/projects/[id=idNumber]/rebuild/+server.ts]: A product also needs to have a ProductDefinition with a RebuildWorkflow specified in order for it to be rebuilt. For the time being, `doProductAction` does make sure this is the case though (it doesn't throw any errors on that, just silently passes it by).

src/routes/(unauthenticated)/api/projects/[id=idNumber]/+server.ts
```diff
@@ -1,7 +1,38 @@
...
+    }
+  });
+
+  if (!project) {
+    return json({ error: `Project id=${projectId} not found` }, { status: 404 });
+  }
+
+  const isOwner = userId === project.Owner.Id;
+  const rebuildableProducts = project.Products.filter(
+    (p) => p.DatePublished && p.PublishLink && !p.WorkflowInstance
```
Comment (FyreByrd) [inline src/routes/(unauthenticated)/api/projects/[id=idNumber]/+server.ts]: All of these filters could be moved to the database query. In this case you could also alter the query to return a `_count` of the products matching the criteria, instead of a list of products.

PR #1363, opened 2025-10-28
Switch org for new project when selected in sidebar

FyreByrd:
Selecting an organization from the sidebar will now cause the organization for a new/import project to be switched in the UI.
This will reset the form inputs, but that should be fine. This is more of a sanity check than anything.

## Summary by CodeRabbit

* **New Features**
  * Added a reusable Organization table for selecting organizations (name, logo, contact).

* **Enhancements**
  * Replaced manual lists with the table across project import and creation flows for consistent selection.
  * Selection now triggers navigation and automatic sync/redirect so the active organization is reflected in routes.

* **Bug Fixes**
  * Safer URL construction when organization ID is missing to avoid malformed links.


chrisvire (approved):
👍

PR #1424, opened 2026-01-06
Add default config for number of jobs to keep

FyreByrd:
On December 19, Scriptoria crashed with an out of memory error from Valkey.
As a temporary measure, the instance size used for Valkey was given increased memory.
This PR introduces changes to limit the number of completed BullMQ jobs that Valkey will keep track of, which will ultimately reduce the amount of memory required.

As of the PR being opened, production has 110k completed jobs recorded across 10 queues, this PR will limit each queue to 3k (1k complete, 2k failed) recorded jobs each, for a maximum of two weeks. This should be sufficient for all our purposes and also little enough that we can downgrade the instance size to what it was before (cost reduction).
Screenshot 2026-01-06 at 11 13 01 AM

## Summary by CodeRabbit

* **Chores**
  * Updated queue cleanup defaults: completed jobs are automatically removed after 2 weeks (or 1000 retained operations) and failed jobs are removed after 2 weeks (or 2000 retained operations).

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/src/lib/server/bullmq/queues.ts
+++ b/src/lib/server/bullmq/queues.ts
@@ -107,11 +107,17 @@ export const getQueueConfig = () => {
     prefix: 'scriptoria',
     telemetry: new BullMQOtel('scriptoria'),
     defaultJobOptions: {
-      // default # jobs to keep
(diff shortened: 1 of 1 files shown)


chrisvire (approved):
👍

PR #1458, opened 2026-02-17
Show previous task comment in individual task page

FyreByrd:
Screenshot 2026-02-17 at 1 34 44 PM

## Summary by CodeRabbit

* **New Features**
  * Show previous task comment on the task page when available to provide context.

* **Style**
  * Task comment component accepts optional class overrides and merges them with default spacing.
  * Form input layout uses conditional styling for improved presentation.

* **Localization**
  * Added "Previous Comment ({activityName})" translations for English, Spanish, and French.

### DIFF
--- a/src/routes/(authenticated)/tasks/[product_id]/+page.svelte
+++ b/src/routes/(authenticated)/tasks/[product_id]/+page.svelte
@@ -163,7 +163,7 @@
     {waiting ? 'Waiting' : data.taskTitle}
   </h2>
   {#if !waiting}
-    {#if data.previousTask}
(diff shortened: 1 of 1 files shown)


FyreByrd:
> Could we add a label to the output? What do you think it could be?

Would "Comment" be sufficient, especially since it's in a visually separate section from the comment box?

FyreByrd:
> @FyreByrd, I think we should hide the field if it is empty. What do you think?
> 

Yeah. Good catch. I thought I had handled that, but apparently I didn't...

PR #1474, opened 2026-03-09
Improve SSE performance

FyreByrd:
Changes:
- Reduce packet size of UserTasks SSE by 30% (based on rough test)
  - Use lookup map for ProjectName and ProductDefinitionName (may not be necessary due to devalue)
  - Minify object keys
- Reduct packet size of Project Details by 50% (based on rough test)
  - Split into four separate sets:
    - Project details: Name, Settings, Owner, etc.
    - Group data: possible owners, possible groups, authors, etc.
    - Org data: stores, product definitions
    - Product data: products, product transitions, etc.
  - Splitting into different sets reduce unnecessary duplication across multiple queries (and allowed logical addition of some SSE triggers)
  - Minified product data
    - Reduction was only 5% before minification

## Summary by CodeRabbit

* **New Features**
  * Added real-time updates for project groups, organizations, and product data through Server-Sent Events.
  * Enhanced product details display with improved data visualization.

* **Improvements**
  * Refined organization and store management interfaces with clearer product associations.
  * Updated task display with improved data structure and rendering.

### DIFF
--- a/src/lib/bullmq.ts
+++ b/src/lib/bullmq.ts
@@ -53,6 +53,7 @@ export enum JobType {
   // Svelte Project SSE
   SvelteSSE_UpdateProject = 'Update Project',
   SvelteSSE_UpdateProjectGroups = 'Update Project Groups',
+  SvelteSSE_UpdateProjectOrg = 'Update Project Org',
   SvelteSSE_UpdateUserTasks = 'Update UserTasks'
 }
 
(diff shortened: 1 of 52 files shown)


7dev7urandom:
This looks good. I particularly like the key minification given the large benefits, and splitting up the project data messages allows for smaller/more targeted future updates. 

You're correct that the deduplication is already handled by devalue, but it doesn't particularly hurt. I haven't tested this myself but I like the changes I see.

Issue #1251, opened 2025-09-08
UDM: Verify Email


chrisvire:
There will be an email verification page where the user will enter the verification code sent through email

PR #1476, opened 2026-03-09
User Data Management: Frontend Pages

ThePitter:
Add pages for user data deletion requests, including sending an email to the user and verifying that it is a real email account. This is separate from the Cloudfare Turnstile integration that will complete this project.
#1251
@James-Voight

## Summary by CodeRabbit

* **New Features**
  * Email-based deletion request and verification flow with 6-digit confirmation codes.
  * New product "Manage my data" pages: deletion request form, verification page, and About view.
  * Cloudflare Turnstile anti-bot verification integrated into deletion flow.
  * App-themed UI with dynamic color derived from app icons.
  * Localized UI copy added for English, Spanish (es-419) and French.

* **Infrastructure**
  * Scheduled cleanup job to remove expired verification requests.


README.md
FyreByrd: This file did not need to be reformatted.

FyreByrd (changes requested):
All of the coderabbit feedback seems relevant to me and should also be addressed.

There is a lot of duplicated code that could be moved to shared files.

If you're pulling significant amounts of code from outside (i.e. the color conversion stuff) please leave a comment crediting where you got it from.

We use a forms library for pretty much every page action handler. Please use it for consistency.

src/routes/(unauthenticated)/products/[id]/confirm/+page.server.ts
```diff
@@ -0,0 +1,238 @@
...
+          Name: true,
+          Organization: { select: { Name: true } }
+        }
+      }
+    }
+  });
+
+  if (!product) throw error(404);
+
+  const developer =
```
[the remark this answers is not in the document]
Comment (ThePitter) [reply src/routes/(unauthenticated)/products/[id]/confirm/+page.server.ts]: @FyreByrd I'm not quite tracking what you are saying here. We should be keeping names internal, but this page is supposed to adapt to the product's name and information. The field in 1440 appears to address what store (like google play) the product is paired with, but I think our intention was when a user clicks the User Data Management link in google play, it shows a page that has the appearance of the app page they were just on (generic pages are not approved by Google). Should we be pulling PackageName or some other value to get the application display name and developer display name?

src/routes/(unauthenticated)/products/[id]/user-data/+layout.server.ts
```diff
@@ -0,0 +1,255 @@
...
+  let manifest: PlayListingManifest | null = null;
+  try {
+    manifest = (await fetch(manifestArtifact.Url).then((r) => r.json())) as PlayListingManifest;
+  } catch {
+    manifest = null;
+  }
+  if (!manifest) return { app };
+
+  // Used to re-home URLs if the underlying storage hostname/bucket changes.
+  const artifactUrl = new URL(manifestArtifact.Url);
```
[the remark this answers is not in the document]
Comment (ThePitter) [reply src/routes/(unauthenticated)/products/[id]/user-data/+layout.server.ts]: @James-Voight

ThePitter:
@James-Voight When you get the chance please review the changes and if you could take a look at the stuff I tagged you on
Let me know if you need anything :D
Once we address the remaining issues we can ask for a review again

src/routes/(unauthenticated)/products/[id]/user-data/+layout.server.ts
```diff
@@ -0,0 +1,255 @@
...
+  return available.at(0) ?? null;
+}
+
+/**
+ * Find a language-specific file path in the manifest file list.
+ * Manifests usually use "<lang>/<file>.txt" but we keep the search flexible.
+ */
+function findFilePath(files: string[], language: string, filename: string) {
+  // Most manifests store entries like: "<lang>/<file>.txt"
+  const exact = new RegExp(`(^|/)${escapeRegExp(language)}/${escapeRegExp(filename)}$`);
```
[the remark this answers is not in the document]
Comment (James-Voight) [reply src/routes/(unauthenticated)/products/[id]/user-data/+layout.server.ts]: I can change this to a pure string/path comparison and get rid of the regex entirely if you want. This lets regex characters to be treated as plain text so that matching stays correct

src/routes/(unauthenticated)/products/[id]/user-data/+layout.server.ts
```diff
@@ -0,0 +1,255 @@
...
+    id: product.Id,
+    icon: null,
+    name: product.Project.Name ?? 'App',
+    developer,
+    themeColor: null,
+    shortDesc: '',
+    longDesc: ''
+  };
+
+  // Prefer published manifest; fall back to latest built manifest if not published yet.
```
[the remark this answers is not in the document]
Comment (James-Voight) [reply src/routes/(unauthenticated)/products/[id]/user-data/+layout.server.ts]: There was a long time where my builds weren't working and fallbacks were needed (no accessible play-listing-manifest). The fallback only keeps branding/content visible when a published manifest doesn’t exist yet. If we want, I can remove the fallback and let unpublished products use defaults instead.

src/routes/(unauthenticated)/+layout.svelte
```diff
@@ -1,16 +1,23 @@
...
+  let { data, children }: Props = $props();
+
+  $effect(() => {
+    l10nMap.value = createl10nMapFromEntries(data.localizedNames);
+  });
 </script>
 
 <div
   data-theme="light"
-  class="grid w-full h-full place-items-center place-content-center [background-color:#0068a6]"
```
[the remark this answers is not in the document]
FyreByrd: I don't think you looked hard enough. It's right here in the Svelte documentation: https://svelte.dev/docs/kit/advanced-routing#Advanced-layouts-Breaking-out-of-layouts

src/routes/(unauthenticated)/products/[id]/confirm/+page.svelte
```diff
@@ -0,0 +1,348 @@
...
+    });
+  }
+
+  let primaryHSL = $derived(hexToDaisyHSL(themeColor));
+  let primaryContentHSL = $derived(hexToDaisyHSL(getReadableTextHex(themeColor)));
+  let primaryHex = $derived(themeColor);
+  let primaryContentHex = $derived(getReadableTextHex(themeColor));
+  let lightBgHex = $derived(lightenColor(themeColor, 0.92));
+
+  $effect(() => {
```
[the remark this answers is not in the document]
Comment (FyreByrd) [reply src/routes/(unauthenticated)/products/[id]/confirm/+page.svelte]: I think it would be better to define these on the `<div>` element. In general `$effect` should be used sparingly and only when some other approach would not work. I would also recommend seeing if it would work to only do the theme change within a root `+layout.svelte` for the UDM route (i.e. `/products/[id]/+layout.svelte`)

static/placeholder-icon.svg
FyreByrd, line 1: I might be able to find a better default icon from one of our internal repositories...

src/routes/(unauthenticated)/products/[id]/confirm/+page.svelte
```diff
@@ -0,0 +1,223 @@
...
+    errors: sendErrors
+  } = superForm(data.sendCodeForm, {
+    onUpdated: ({ form }) => {
+      if (form.valid && form.message?.step === 'verify') {
+        step = 'code';
+        $sendForm.email = form.message.email;
+        $verifyForm.email = form.message.email;
+        sessionStorage.setItem(confirmEmailStorageKey, form.message.email);
+      } else if (!form.valid && !form.message?.error) {
+        $sendMessage = { error: m.udm_confirm_send_code_failed() };
```
Comment (FyreByrd) [inline src/routes/(unauthenticated)/products/[id]/confirm/+page.svelte]: Most of these i18n keys are missing?

src/routes/(unauthenticated)/products/[id]/+layout.svelte
Comment (FyreByrd) [inline src/routes/(unauthenticated)/products/[id]/+layout.svelte:1]: This route will probably cause routing conflicts with changes I made in #1478. Maybe restructure this folder from `/products/[id]/*` to `/user-data/[product_id]/*`

src/lib/products/UDMtypes.ts
```diff
@@ -0,0 +1,18 @@
+export type PlayListingManifest = {
+  url: string; // Base URL prefix where the listing files can be fetched from.
+  icon: string; // Icon path (or URL) inside the listing bundle.
+  color: string; // Brand color hex (e.g. "#1563ff")
+  'default-language'?: string; // Default language tag for the listing bundle (e.g. "en-US").
+  languages?: string[]; // Languages included in the bundle (language tags).
+  files?: string[]; // Paths to files within the bundle (usually "<lang>/<file>.txt").
+};
+
+export type AppInfo = {
```
FyreByrd: I really don't think this type is at all necessary. You could definitely simplify the two load functions that use it.

src/lib/products/UDMtypes.ts
```diff
@@ -0,0 +1,18 @@
+export type PlayListingManifest = {
```
FyreByrd: Rather than having this as a separate type that you have to define, you could instead define this as the return type of a function that handles the manifest. In fact, I already have a function in `/downloads/apk/*` that you could reuse and maybe relocate.

src/lib/server/job-executors/databaseCleanup.ts
FyreByrd, line 1: Rather than having this as an entirely separate job namespace, I think this would be more useful as one of the many jobs available in System Recurring, with the possibility of adding more database records to clean up in the future. I would also recommend changing the names accordingly. Please take a look at the other system recurring jobs (especially what the migrate job looked like historically)

src/routes/(unauthenticated)/api/delete-request/+server.ts
FyreByrd, line 1: We already have a validation library called `valibot`. Please use it for the request validation where possible.

src/routes/(unauthenticated)/api/delete-request/+server.ts
```diff
@@ -0,0 +1,218 @@
...
+import { DatabaseWrites } from '$lib/server/database';
+import prisma from '$lib/server/database/prisma';
+import { sendEmail } from '$lib/server/email-service/EmailClient';
+import { EmailLayoutTemplate, addProperties } from '$lib/server/email-service/EmailTemplates';
+
+const EMAIL_PATTERN = /^[^\s@]+@[^\s@]+\.[^\s@]+$/;
+const UUID_PATTERN = /^[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/i;
+const TURNSTILE_TIMEOUT_MS = 5000;
+
+interface DeleteRequestBody {
```
FyreByrd: If you use `valibot` for validation, you won't need to define this type

src/lib/server/job-executors/databaseCleanup.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export async function databaseCleanupWorker(
+  job: Job<BullMQ.UserDataManagement.Cleanup>
+): Promise<void> {
+  // This job cleans up old email verification requests that are past their expiration date. It runs every night.
+
+  // TODO: Determine if we want to only delete instances where the DateConfirmed is null. If we use this record to track confirmed requests that haven't actually had their data deleted yet, we might want to keep the record until the data is deleted, even if it's past the expiration date. For now, we'll assume that once it's past the expiration date, we can delete it regardless of whether it was confirmed or not.
+  await DatabaseWrites.productUserChanges.deleteMany({
+    where: {
+      DateExpires: {
```
FyreByrd, line 9: Make sure this doesn't delete confirmed requests.

src/lib/utils/theme.ts
```diff
@@ -0,0 +1,174 @@
...
+}
+
+export function mixColors(hexA: string, hexB: string, amount = 0.5) {
+  const a = hexToRgb(hexA);
+  const b = hexToRgb(hexB);
+  const mix = (from: number, to: number) => Math.round(from + (to - from) * amount);
+  return rgbToHex(mix(a.r, b.r), mix(a.g, b.g), mix(a.b, b.b));
+}
+
+export function hexToDaisyHSL(hex: string) {
```
FyreByrd: Bikeshedding here, but is there anything DaisyUI-related about this function? Would it not be better names as `hexToHSL`?

src/lib/utils/theme.ts
```diff
@@ -0,0 +1,174 @@
...
+        h = (r - g) / d + 4;
+        break;
+    }
+    h /= 6;
+  }
+
+  return `${(h * 360).toFixed(1)} ${(s * 100).toFixed(1)}% ${(l * 100).toFixed(1)}%`;
+}
+
+export function deriveColorFromIcon(iconSrc: string, fallbackColor: string): Promise<string> {
```
FyreByrd: Can I get an explanatory comment on *how* this function derives a color from an icon image?

FyreByrd (changes requested):
This is the feedback I have for now. I will be working on having a separate locale implementation for UDM.

src/lib/locales/en-US.json
```diff
@@ -602,5 +578,33 @@
...
+  "udm_email_placeholder": "you@example.com",
+  "udm_email_help": "Use the email associated with your account.",
+  "udm_scope_label": "Deletion Scope",
+  "udm_scope_data_title": "Delete my data",
+  "udm_scope_data_desc": "Your login remains active, but your personal content will be permanently deleted.",
+  "udm_scope_account_title": "Delete my account and all associated data",
+  "udm_scope_account_desc": "This will permanently remove your login and saved content.",
+  "udm_scope_warning": "Warning: Deletions are permanent and cannot be undone. Some data may be retained for legal or compliance purposes.",
+  "udm_items_removed_title": "Items to be removed",
+  "udm_item_bookmarks": "Bookmarks",
```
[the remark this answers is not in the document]
ThePitter: I'll implement it now. My comment wasn't so much about safety (since that would require someone editing this file to have malicious intent, which is unlikely) as much as it is about I ran across a comment that industry standard of localization discourages having html tags because not every language will format the same way.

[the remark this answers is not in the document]
ThePitter: Second note, lint is saying that there is a 'no at html tags' rule enabled. Should I have it ignore that on this page? @FyreByrd

src/lib/products/UDMtypes.ts
ThePitter: Could you provide more detail into what you are envisioning here? Based on the analysis that I've done between the `downloads/apk/page@.svelte` and `user-data/layout.server.ts` (and others) which utilize PlayListingManifest, we could use the _getManifest function, but that will *increase* the number of requests to get the right data than the way it is currently implemented (it requests only the language needed, `_getManifest` queries all languages every time). I've moved the types directly into the functions being queried in the two user-data files and removed UDMTypes. This should keep the function calls efficient while removing the unassociated type function.

src/lib/products/UDMtypes.ts
ThePitter: This has been removed and put directly into the functions for simplicity.

src/lib/server/job-executors/databaseCleanup.ts
ThePitter: Moved logic from `databaseCleanup` to `system.ts` and renamed the job to `databaseCleanup` so other tasks could be added in the future. The job type is now `System_DatabaseCleanup` and the job scheduler ID is now `System_DatabaseCleanup`.

src/lib/utils/theme.ts
ThePitter: Fair point. I think technically the format is CSS, not HSL. I renamed the function.

src/lib/utils/theme.ts
ThePitter: While @James-Voight could probably provide a better answer since he wrote it, here's my analysis (aided by @gemini_cli) I also added a JSDoc comment that provides an explanation. Basically though, this function loads the icon in an off-screen HTML5 canvas, iterates through non-transparent pixels, calculates the average color of all pixels, then returns the Hex code for that color.

ThePitter: I adjusted the function to draw onto a 64x64 canvas so the image is averaged down by the browser and the final pixel average only requires a small number of iterations to find, instead of the native size.

src/routes/(unauthenticated)/api/delete-request/+server.ts
ThePitter: resolved.

src/routes/(unauthenticated)/products/[id]/confirm/+page.server.ts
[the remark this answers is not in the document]
Comment (ThePitter) [reply src/routes/(unauthenticated)/products/[id]/confirm/+page.server.ts]: Ah, I see now. Fixed. Requisition Store's GooglePlayTitle instead of Project's Name and Organization.

src/routes/(unauthenticated)/products/[id]/confirm/+page.svelte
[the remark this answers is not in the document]
Comment (ThePitter) [reply src/routes/(unauthenticated)/products/[id]/confirm/+page.svelte]: This worked out great! I was able to move the remaining duplicate logic on the UDM pages related to styling calculations to one unified `+layout.svelte` file.

src/routes/(unauthenticated)/products/[id]/confirm/+page.svelte
Comment (ThePitter) [reply src/routes/(unauthenticated)/products/[id]/confirm/+page.svelte]: That's odd. They're there for me at the bottom. The one at line 38 is at line 636 in the locales.

src/routes/(unauthenticated)/+layout.svelte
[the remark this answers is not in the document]
ThePitter: Woahhhhhhh I have literally looked for a way to do that in two separate projects and haven't seen that before. That's cool! The changes should have been already reverted and the updated layout is put into a more narrow-scoped layout file.

static/placeholder-icon.svg
ThePitter: Just let us know!

ThePitter:
Whoops created a merge commit, undid that. There are conflicts though that need to be resolved. Our develop needs to be fast-forwarded first though

ThePitter:
There's got to be a better way to resolve conflicts when rebasing then having to force push every time. @judah-sotomayor ?

FyreByrd:
> There's got to be a better way to resolve conflicts when rebasing then having to force push every time.

Unfortunately, that's just the way it is. Rebasing will always require a force-push.

src/lib/locales/en-US.json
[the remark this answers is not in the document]
FyreByrd: Sorry for the delay, no clue how I forgot to respond. Yes, that's what would be necessary.

paraglide/default.inlang/cache/plugins/2sy648wh9sugi
FyreByrd, line 1: This file should not be commited. Paraglide should have generated a `.gitignore` file.

paraglide/default.inlang/cache/plugins/ygx0uiahq6uw
FyreByrd, line 1: This file should not be commited. Paraglide should have generated a `.gitignore` file.

paraglide/default.inlang/project_id
FyreByrd, line 1: This file should not be commited. Paraglide should have generated a `.gitignore` file.

paraglide/google-play.inlang/cache/plugins/2sy648wh9sugi
FyreByrd, line 1: This file should not be commited. Paraglide should have generated a `.gitignore` file.

paraglide/google-play.inlang/cache/plugins/ygx0uiahq6uw
FyreByrd, line 1: This file should not be commited. Paraglide should have generated a `.gitignore` file.

paraglide/google-play.inlang/project_id
FyreByrd, line 1: This file should not be commited. Paraglide should have generated a `.gitignore` file.

project.inlang/cache/plugins/2sy648wh9sugi
FyreByrd, line 1: This file should not be commited. Paraglide should have generated a `.gitignore` file.

project.inlang/cache/plugins/ygx0uiahq6uw
FyreByrd, line 1: This file should not be commited. Paraglide should have generated a `.gitignore` file.

src/lib/google-play/components/LocaleSelector.svelte
```diff
@@ -76,20 +76,13 @@
         {#each locales.toSorted( (a, b) => byString(displayNames.get(a)?.display, displayNames.get(b)?.display, current) ) as locale}
           {@const { display, native, fallback } = displayNames.get(locale)!}
           <li class="w-full">
-            <div
+            <button
```
FyreByrd, line 79: I don't remember why I forgot to do it this way instead, but it is probably better. Thanks.

src/lib/locales/en-US.json
FyreByrd, line 1: The UDM messages should be removed from here, since they are present in the GooglePlay locales.

src/lib/locales/es-419.json
FyreByrd, line 1: The UDM messages should be removed from here, since they are present in the GooglePlay locales.

src/lib/locales/fr-FR.json
FyreByrd, line 1: The UDM messages should be removed from here, since they are present in the GooglePlay locales.

src/lib/products/UDMtypes.ts
FyreByrd: This type is obsolete and can be removed.

src/lib/products/UDMtypes.ts
FyreByrd: I really don't think you need this type at all. You only use it in one place.

src/lib/server/bullmq/types.ts
```diff
@@ -296,6 +296,14 @@ export namespace SvelteProjectSSE {
   }
 }
 
+// User Data Management: jobs related to cleanup of user data, such as old email verification requests, GDPR data, etc.
+export namespace UserDataManagement {
+  // This cleans up old email verification requests that are past their expiration date. It runs every night.
+  export interface Cleanup extends BaseJob {
```
FyreByrd: Please move this job to the `System` namespace.

src/lib/server/bullmq/BullWorker.ts
```diff
@@ -110,6 +110,19 @@ export class SystemRecurring<J extends BullMQ.RecurringJob> extends BullWorker<J
...
       }
     );
+    getQueues().SystemRecurring.upsertJobScheduler(
+      BullMQ.JobSchedulerId.CleanupUserData,
+      {
+        pattern: '0 0 * * *', // every night at midnight
+        immediately: false
+      },
+      {
+        name: 'Cleanup User Data',
```
FyreByrd: If we're making this more generic, you could probably change it to something along the lines of `Cleanup Old/Expired Data`.

src/lib/server/job-executors/databaseCleanup.ts
```diff
@@ -0,0 +1,18 @@
+import type { Job } from 'bullmq';
+import type { BullMQ } from '../bullmq';
+import { DatabaseWrites } from '../database';
+
+export async function databaseCleanupWorker(
```
FyreByrd: Please move this function to either `job-executors/system/index.ts` or `job-executors/system/cleanup.ts`

FyreByrd: This name is unnecessarily verbose, you could probably just name it `cleanup`

src/lib/bullmq.ts
```diff
@@ -52,12 +52,15 @@ export enum JobType {
   Email_ProjectImportReport = 'Project Import Report',
   // Svelte Project SSE
   SvelteSSE_UpdateProject = 'Update Project',
-  SvelteSSE_UpdateUserTasks = 'Update UserTasks'
+  SvelteSSE_UpdateUserTasks = 'Update UserTasks',
+  // User Data Management
+  UDM_Cleanup = 'User Data Management Cleanup'
```
FyreByrd: Please rename this to `System_Cleanup`

src/lib/bullmq.ts
```diff
@@ -52,12 +52,15 @@ export enum JobType {
...
 }
 
 export enum JobSchedulerId {
   SystemStatusEmail = 'SystemStatusEmail',
   RefreshLangTags = 'RefreshLangTags',
   CheckSystemStatuses = 'CheckSystemStatuses',
-  MigrateChunks = 'MigrateChunks'
+  MigrateChunks = 'MigrateChunks',
+  CleanupUserData = 'CleanupEmailVerificationRequests'
```
FyreByrd: Please rename this to `CleanupExpiredData`

src/lib/utils/theme.ts
FyreByrd: This really shouldn't be necessary at all. We don't need to go through all the extra effort to derive a color from the app icon when we already have a perfectly usable color from the app manifest.

static/placeholder-icon.svg
FyreByrd: I found some icons and sent them to you over discord. Please take a look at `$lib/icons/index.ts` Lines 30-41 for how I've done something similar with icons specific to the different app types.

src/routes/(unauthenticated)/(google-play)/user-data/[productId]/about/+page.svelte
```diff
@@ -0,0 +1,111 @@
...
+
+  let { data }: Props = $props();
+
+  const app = data.app;
+  const currentLocale = data.locale as Locale;
+  const iconSrc = app.icon ?? DEFAULT_ICON;
+  let themeColor = $state(app.themeColor ?? '#0e795b');
+
+  let themeVars = $derived(getThemeVariables(themeColor));
+  let themeStyle = $derived(
```
Comment (FyreByrd) [inline src/routes/(unauthenticated)/(google-play)/user-data/[productId]/about/+page.svelte]: This is doing the same thing as `applyThemeToDocument` please choose one or the other.

src/lib/utils/theme.ts
```diff
@@ -0,0 +1,174 @@
...
+    lightBgHex: lightenColor(themeColor, 0.92),
+    darkBgHex: mixColors(themeColor, '#0b1220', 0.88),
+    shellLightHex: mixColors(themeColor, '#ffffff', 0.93),
+    shellDarkHex: mixColors(themeColor, '#111827', 0.84),
+    headerLightHex: mixColors(themeColor, '#ffffff', 0.85),
+    headerDarkHex: mixColors(themeColor, '#0f172a', 0.72)
+  };
+}
+
+export function applyThemeToDocument(themeColor: string) {
```
FyreByrd: Instead of applying this directly to the root document element, why not do this as a string that could be passed into the `style` prop of the container element.

src/routes/(unauthenticated)/(google-play)/user-data/about/+page.svelte
```diff
@@ -0,0 +1,111 @@
...
+  });
+
+  onMount(() => {
+    if (!app.themeColor) {
+      deriveColorFromIcon(iconSrc, themeColor).then((hex) => (themeColor = hex));
+    }
+  });
+</script>
+
+<svelte:head>
```
FyreByrd: This styling shouldn't be necessary??

src/lib/utils/theme.ts
```diff
@@ -0,0 +1,174 @@
...
+  if (typeof document === 'undefined') return;
+  const vars = getThemeVariables(themeColor);
+  const root = document.documentElement;
+  root.style.setProperty('--color-primary', vars.primaryHex);
+  root.style.setProperty('--color-primary-content', vars.primaryContentHex);
+  root.style.setProperty('--p', vars.primaryHSL);
```
FyreByrd: I'm fairly certain that `--p`, `--pf`, and `--pc` are not at all necessary.

src/lib/utils/theme.ts
FyreByrd, line 1: The vast majority of this functionality is not necessary. You'd probably be better served by creating a CSS stylesheet and using some built in CSS functionality like [`color-mix()`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/color_value/color-mix). If you use that, you should only need one function to parse a RGB hex string and convert that to HSL.

src/routes/(unauthenticated)/(google-play)/user-data/about/+page.svelte
```diff
@@ -0,0 +1,111 @@
...
+  } from '$lib/utils/theme';
+
+  interface Props {
+    data: PageData;
+  }
+
+  let { data }: Props = $props();
+
+  const app = data.app;
+  const currentLocale = data.locale as Locale;
```
FyreByrd: Paraglide has a function called `getLocale` that you should use instead. `data.locale` is meant for server side use and I had only been sending it to the client in the first place for debugging purposes when I was working on the separate locales feature.

src/routes/(unauthenticated)/(google-play)/user-data/about/+page.svelte
```diff
@@ -0,0 +1,111 @@
...
+              {app.longDesc}
+            </div>
+          </details>
+        </div>
+      </div>
+
+      <a
+        class="btn w-full mt-6 border border-primary/20 text-black shadow-sm dark:text-white"
+        href={localizeHref(`/user-data?${data.udmQuery}`, { locale: currentLocale })}
+        style="background-color: var(--color-primary);"
```
FyreByrd: `class="bg-primary` is equivalent to `style="background-color: var(--color-primary)` but much shorter...

FyreByrd: In fact, since you're already using `class="btn"` you could do even better by just adding `btn-primary` to the class-list. I think the class here could be simplified all the way down to `btn btn-primary mt-6`. You really shouldn't need anything else.

src/lib/utils/theme.ts
```diff
@@ -0,0 +1,174 @@
...
+        resolve(fallbackColor);
+      }
+    };
+
+    img.onerror = () => resolve(fallbackColor);
+    img.src = iconSrc;
+  });
+}
+
+export function getThemeVariables(themeColor: string) {
```
FyreByrd: If you have a separate stylesheet with all the math implemented using CSS `color-mix()` you should only need to be overwriting `--color-primary` and `--color-primary-content`

FyreByrd: Depending on how comfortable you are with CSS, you may even be able to figure out the value of `--color-primary-content` using just CSS.

src/lib/utils/theme.ts
FyreByrd: Or, better yet, have a separate stylesheet and you only need to pass in one or two vars total.

src/routes/(unauthenticated)/(google-play)/user-data/[productId]/confirm/+page.server.ts
Comment (FyreByrd) [inline src/routes/(unauthenticated)/(google-play)/user-data/[productId]/confirm/+page.server.ts:1]: All of your i18n message calls will need to have the locale passed in manually so they work well on the server.

src/routes/(unauthenticated)/(google-play)/user-data/confirm/+page.server.ts
```diff
@@ -0,0 +1,210 @@
...
+import type { Actions, PageServerLoad } from './$types';
+import { m } from '$lib/google-play/paraglide/messages';
+import { DatabaseReads, DatabaseWrites } from '$lib/server/database';
+import prisma from '$lib/server/database/prisma';
+import { sendEmail } from '$lib/server/email-service/EmailClient';
+
+function createSchemas() {
+  const uuidSchema = v.pipe(
+    v.string(),
+    v.regex(
```
FyreByrd: This is **completely unnecessary**. If you actually take even a minute to look at the valibot docs, there is already a pre-defined UUID validation action that you can use. Please use that rather than a regex you haven't even bothered to properly attribute.

src/routes/(unauthenticated)/(google-play)/user-data/confirm/+page.server.ts
```diff
@@ -0,0 +1,210 @@
...
+        }
+      );
+
+      await sendEmail(
+        [{ email: normalizedEmail, name: normalizedEmail }],
+        m.udm_email_subject(),
+        m.udm_email_body({ code })
+      );
+
+      if (process.env.NODE_ENV === 'development') {
```
FyreByrd: This is **completely unnecessary**. `sendEmail` *already handles this*.

src/routes/(unauthenticated)/(google-play)/user-data/confirm/+page.svelte
```diff
@@ -0,0 +1,225 @@
...
+      if (form.valid && form.message?.verified) {
+        step = 'verified';
+        sessionStorage.removeItem(confirmEmailStorageKey);
+      } else if (!form.valid && !form.message?.error) {
+        $verifyMessage = { error: m.udm_error_invalid_code_retry() };
+      }
+    }
+  });
+
+  let step = $state(
```
FyreByrd: VS Code is showing this as having type `string`. You may want to tighten this slighly.

src/routes/(unauthenticated)/(google-play)/user-data/confirm/+page.server.ts
```diff
@@ -0,0 +1,210 @@
...
+                DateExpires: now
+              }
+            });
+          }
+
+          if (latestRequest) {
+            await tx.productUserChanges.update({
+              where: { Id: latestRequest.Id },
+              data: {
+                Change: m.udm_change_description(),
```
FyreByrd: I don't think this is what you want to put in the `Change` field...

src/routes/(unauthenticated)/(google-play)/user-data/+layout.server.ts
```diff
@@ -0,0 +1,89 @@
...
+
+function getSearchParam(url: URL, keys: string[]) {
+  for (const key of keys) {
+    const value = url.searchParams.get(key)?.trim();
+    if (value) return value;
+  }
+  return '';
+}
+
+export const load: LayoutServerLoad = async ({ locals, url }) => {
```
FyreByrd: Per conversation with Chris:
1. This route should only use the product id. It should not use the package name. This is to reduce guessability, since it is a publicly available route.
2. The product id should be part of the route itself, not a URL search param.
3. If you want to verify that it is a valid UUID, please use valibot instead of rolling your own regex validation. But realistically it should just be enough to 404 if the database query fails.

src/routes/(unauthenticated)/(google-play)/user-data/+layout.server.ts
```diff
@@ -0,0 +1,89 @@
...
+          Organization: { select: { Name: true } }
+        }
+      }
+    }
+  });
+
+  if (!product) throw error(404);
+
+  const canonicalPackage = queryPackage || product.PackageName || '';
+  const developer =
```
FyreByrd: As we've already discussed *multiple* times at this point, this is supposed to be `Store.GooglePlayTitle`. Anything other than that or your `'Unknown developer' fallback is a possible **data leak** and therefore unacceptable.

src/routes/(unauthenticated)/(google-play)/user-data/+page.server.ts
FyreByrd, line 1: Why did you delete this file?

src/routes/(unauthenticated)/(google-play)/user-data/+page.svelte
```diff
@@ -1,67 +1,312 @@
...
-            ? tryLocalize(data.l10nMap, locale, 'languages', altLocale, fallback)
-            : native
-          : native;
-        return [locale, { display: displayLocal, native: displayNative, fallback }];
-      })
-    )
+  const app = data.app;
+  const iconSrc = app.icon ?? DEFAULT_ICON;
+  const currentLocale = data.locale as Locale;
+  const localeFallbacks = $derived(
```
FyreByrd: If you want a non-null value for this, please do this on the server side where fallbacks originates from.

src/routes/(unauthenticated)/(google-play)/user-data/+page.svelte
```diff
@@ -1,67 +1,312 @@
...
+    new Map(Array.from(data.fallbacks.entries()).map(([key, value]) => [key, value ?? key]))
   );
+  const confirmEmailStorageKey = `udm-confirm-email:${app.id}`;
+  let themeColor = $state(app.themeColor ?? '#0e795b');
+
+  async function submitForm() {
```
FyreByrd: Most of this functionality would be better as form actions using the superforms library.

src/routes/(unauthenticated)/(google-play)/user-data/+page.svelte
```diff
@@ -1,67 +1,312 @@
...
+              <span
+                class="label-text text-xs font-bold uppercase tracking-wide text-base-content/70"
+              >
+                {m.udm_verification_label()}
+              </span>
+            </p>
+            <div class="mt-2">
+              <div
+                class="cf-turnstile"
+                data-sitekey="0x4AAAAAACW5LHZjYD-EBWMs"
```
FyreByrd: Is this something that should be user-facing? Just curious

src/routes/(unauthenticated)/(google-play)/+layout@.svelte
FyreByrd, line 1: Why did you change this file? I'm pretty sure this was fine as is.

src/routes/(unauthenticated)/api/delete-request/+server.ts
FyreByrd, line 1: Please relocate this code as a form action for the relevant page rather than having it as an API endpoint. This will make the form handling in the one page that cares about it much easier.

src/routes/(unauthenticated)/api/delete-request/+server.ts
```diff
@@ -0,0 +1,218 @@
...
+  email: string;
+  token: string;
+  productId: string;
+  deletionType: 'data' | 'account';
+}
+
+function parseDeleteRequestBody(payload: unknown): DeleteRequestBody | null {
```
FyreByrd: Please use valibot for your validation.

src/hooks.server.ts
```diff
@@ -60,12 +60,7 @@ const paraglideHandle: Handle = ({ event, resolve }) => {
       event.request = localizedRequest;
       return resolve(event, {
         transformPageChunk: ({ html }) => {
-          return html
```
FyreByrd, line 63: Why did you change this? This works fine as is. Please don't make unnecessary changes to the GooglePlay locales code. I spent a moderate amount of time getting this to work correctly.

FyreByrd (changes requested):
This turned out to be a lot more feedback than I anticipated... Sorry about that. Good luck getting this all taken care of before due date.

A few things that I want to take note of:
1. For the theming stuff, it may be better if you have a CSS stylesheet with all the stuff you need there. (e.g. CSS has a built in color mix function).
2. You still have a lot of custom validation and form handling that could be cleaner if you use valibot and sveltekit-superforms.
3. You still haven't done anything to add the user tasks.
4. Please use the product id as the primary way to find the app info, and have it directly in the route instead of as a URL search param.
5. Based on the fact that some files were commited that shouldn't have been, you may have forgotten to upgrade ParaglideJS on your local machines when rebasing off of my GooglePlay locale changes. I should have mentioned that was necessary, so that's partially on me.

src/hooks.server.ts
FyreByrd: The `getTextDirection` function was added in a more recent version of ParaglideJS. If you had issues with this, it may be because you didn't upgrade which version of Paraglide you were using after rebasing off of my GooglePlay locale changes.

src/routes/(unauthenticated)/(google-play)/+layout@.svelte
```diff
@@ -21,15 +21,21 @@
 
   afterNavigate(() => {
     if (browser) {
-      // if page isn't explicitly localized redirect to locale based on preference
-      if (!page.url.pathname.match(`^/${data.locale}/`)) {
```
FyreByrd, line 25: This initial check should have been fine. If a locale is present in the url, and it doesn't match one of the supported locales, then the server will throw an error rather than attempt to render the page. If you're worried about a ReDoS attack, it's actually impossible in this situation, as the locale passed to this regex will **always** be a valid locale, none of which will ever be formatted as a malicious regex.

src/lib/utils/theme.ts
FyreByrd: It is also worth noting that the manifest may have up to three colors in it, separated by newline characters. The first one is the light-mode color, and the other two are sepia and dark-mode, but I don't remember the ordering.

src/routes/(unauthenticated)/(google-play)/user-data/+page.svelte
FyreByrd: For reference, the necessary change should be in `(google-play)/+layout.server.ts`

src/routes/(unauthenticated)/(google-play)/user-data/+page.svelte
James-Voight: Yes, Josiah would know more than me but the Turnstile site key is public and is what allows the widget to render on the page. The sensitive value is the Turnstile secret key, which stays server-side and is only used when verifying tokens with Cloudflare. Basically it's good as long as the verification logic is on the backend.

src/routes/(unauthenticated)/(google-play)/user-data/[productId]/about/+page.svelte
Comment (James-Voight) [reply src/routes/(unauthenticated)/(google-play)/user-data/[productId]/about/+page.svelte]: removed the applyThemeToDocument

src/lib/server/job-executors/system/cleanup.ts
```diff
@@ -0,0 +1,16 @@
+import type { Job } from 'bullmq';
+import type { BullMQ } from '../../bullmq';
+import { DatabaseWrites } from '../../database';
+
+export async function cleanup(_job: Job<BullMQ.System.CleanupExpiredData>): Promise<void> {
+  // This job cleans up old email verification requests that are past their expiration date. It runs every night.
+
+  // TODO: Determine if we want to only delete instances where the DateConfirmed is null. If we use this record to track confirmed requests that haven't actually had their data deleted yet, we might want to keep the record until the data is deleted, even if it's past the expiration date. For now, we'll assume that once it's past the expiration date, we can delete it regardless of whether it was confirmed or not.
+  await DatabaseWrites.productUserChanges.deleteMany({
```
FyreByrd: I'm fairly certain that `DateExpires` is only for tracking if the Confirmation Code has expired. The request itself shouldn't be expired unless confirmation fails.

src/lib/server/bullmq/BullWorker.ts
```diff
@@ -110,6 +110,19 @@ export class SystemRecurring<J extends BullMQ.RecurringJob> extends BullWorker<J
...
       }
     );
+    getQueues().SystemRecurring.upsertJobScheduler(
+      BullMQ.JobSchedulerId.CleanupExpiredData,
+      {
+        pattern: '0 0 * * *', // every night at midnight
+        immediately: false
+      },
+      {
+        name: 'Cleanup old/Expired Data',
```
FyreByrd: I know this is bikeshedding at this point, but please capitalize the `o`.

src/lib/server/job-executors/system/cleanup.ts
```diff
@@ -0,0 +1,16 @@
...
+    where: {
+      DateExpires: {
+        lt: new Date()
+      }
+    }
+  });
```
FyreByrd, line 13: Please add `job.updateProgress(100);` before returning. Please also return the number of records deleted.

src/lib/server/bullmq/BullWorker.ts
```diff
@@ -110,6 +110,19 @@ export class SystemRecurring<J extends BullMQ.RecurringJob> extends BullWorker<J
         }
       }
     );
+    getQueues().SystemRecurring.upsertJobScheduler(
+      BullMQ.JobSchedulerId.CleanupExpiredData,
+      {
+        pattern: '0 0 * * *', // every night at midnight
```
FyreByrd: You don't need to change this, but you could also use `@daily` instead, which has the same effect but won't need an explanatory comment.

src/lib/bullmq.ts
```diff
@@ -52,12 +52,15 @@ export enum JobType {
   Email_ProjectImportReport = 'Project Import Report',
   // Svelte Project SSE
   SvelteSSE_UpdateProject = 'Update Project',
-  SvelteSSE_UpdateUserTasks = 'Update UserTasks'
+  SvelteSSE_UpdateUserTasks = 'Update UserTasks',
+  // User Data Management
+  System_Cleanup = 'System Cleanup'
```
FyreByrd: I know this is probably bikeshedding, but can you please relocate this to be next to the other `System_*` jobs?

.gitignore
```diff
@@ -14,3 +14,9 @@ languages
 /out/
 Thumbs.db
 ehthumbs.db
+
+# Inlang/Paraglide generated files
```
FyreByrd: No. This is not it at all. ParaglideJS, if you are using the proper version, generates **its own** `.gitignore` file within its project directory.
1. Ensure the dev server is **not** running.
2. Revert this change.
3. Delete the files it is supposed to be hiding.
4. `npm i` to update which version of ParaglideJS you have locally.
5. Restart the dev server.
6. The files should all be properly hidden by ParaglideJS' own `.gitignore`.

src/routes/(unauthenticated)/(google-play)/user-data/[productId]/+layout.server.ts
```diff
@@ -0,0 +1,93 @@
...
+import type { LayoutServerLoad } from './$types';
+import { deLocalizeUrl, localizeUrl } from '$lib/google-play/paraglide/runtime';
+import {
+  getLatestManifest,
+  resolveManifestLanguage,
+  translateManifest
+} from '$lib/products/server';
+import { DatabaseReads } from '$lib/server/database';
+
+function getFallbackIcon(appTypeName: string | null | undefined) {
```
Comment (FyreByrd) [inline src/routes/(unauthenticated)/(google-play)/user-data/[productId]/+layout.server.ts]: Why a string comparison? Also, you *completely* ignored my recommendation on how to handle this based on the implementation of `getAppIcon` in `$lib/icons/index.ts`. I would greatly prefer it if you did a near identical implementation, ideally located in `$lib/icons/index.ts`.

src/routes/(unauthenticated)/(google-play)/user-data/[productId]/+layout.server.ts
```diff
@@ -0,0 +1,93 @@
...
+  if (!product) throw error(404);
+
+  const developer = product.Store.GooglePlayTitle?.trim() || 'Unknown developer';
+  const longDesc =
+    translatedManifest['full_description.txt'] || translatedManifest['description.txt'] || '';
+  const fallbackIcon = getFallbackIcon(product.Project.ApplicationType?.Name);
+
+  const app = {
+    id: product.Id,
+    icon: translatedManifest.icon || fallbackIcon,
```
Comment (FyreByrd) [inline src/routes/(unauthenticated)/(google-play)/user-data/[productId]/+layout.server.ts]: The current implementation of `translatedManifest.icon` will never use the fallback, as the icon link also includes the baseUrl of the manifest, therefore the icon string will never be empty.

src/routes/(unauthenticated)/(google-play)/user-data/[productId]/about/+page.svelte
```diff
@@ -0,0 +1,85 @@
...
+  let { data }: Props = $props();
+
+  const app = data.app;
+  const currentLocale = getLocale() as Locale;
+  const iconSrc = app.icon ?? DEFAULT_ICON;
+  let themeColor = $state(app.themeColor ?? '#0e795b');
+
+  let themeVars = $derived(getThemeVariables(themeColor));
+  let themeStyle = $derived(
+    `--color-primary: ${themeVars.primaryHex}; --color-primary-content: ${themeVars.primaryContentHex}; --udm-shell-light: ${themeVars.shellLightHex}; --udm-shell-dark: ${themeVars.shellDarkHex}; --udm-header-light: ${themeVars.headerLightHex}; --udm-header-dark: ${themeVars.headerDarkHex};`
```
Comment (FyreByrd) [inline src/routes/(unauthenticated)/(google-play)/user-data/[productId]/about/+page.svelte]: This string is duplicated in three different locations. Please find a way to make it more DRY

src/routes/(unauthenticated)/(google-play)/user-data/[productId]/confirm/+page.server.ts
Comment (FyreByrd) [reply src/routes/(unauthenticated)/(google-play)/user-data/[productId]/confirm/+page.server.ts:1]: I see that you fixed this for this file, and only this file. The same issue exists elsewhere too.

src/routes/(unauthenticated)/(google-play)/user-data/[productId]/+page.svelte
```diff
@@ -0,0 +1,331 @@
...
+
+  interface Props {
+    data: PageData;
+  }
+
+  let { data }: Props = $props();
+  let turnstileToken: string | null = null;
+
+  const app = data.app;
+  const iconSrc = app.icon ?? DEFAULT_ICON;
```
Comment (FyreByrd) [inline src/routes/(unauthenticated)/(google-play)/user-data/[productId]/+page.svelte]: This should be using the app type specific icon, not the default one.

FyreByrd (changes requested):
I noticed a couple things that needed fixing or more info, so here you go.

A lot of your styling in the UDM pages is a mess and unnecessarily verbose. Some of the inline classes are also out of date for the versions of DaisyUI and TailwindCSS that we are using.

I am also still waiting to see a CSS-only implementation of your theming for the UDM pages. And by CSS-only, I don't mean inline styles. I would prefer an implementation with an external CSS file. Ideally, you would only need to pass in one, maybe two, CSS variable.

.gitignore
James-Voight: how, that is how it is. sorry I am not that familiar with Paraglide. Thank you

src/lib/google-play/udm-theme.css
```diff
@@ -0,0 +1,37 @@
+.udm-theme {
+  /* Product pages set --color-primary from the Google Play manifest. */
```
FyreByrd, line 2: Why are some of these oklch and some srgb?

src/lib/icons/index.ts
```diff
@@ -40,6 +40,19 @@ export function getAppIcon(type: ApplicationType) {
   );
 }
 
+export function getAppFallbackIcon(type: ApplicationType | number | null | undefined) {
```
FyreByrd: This is better, but still isn't quite what I'm looking for.
Here is exactly what I'm looking for:
1. Move the placeholder icons to a `google-play` subfolder of `$lib/icons`, have one for each app type with the pattern `<app-name>.svg`, exactly the same as in `$lib/icons/app-builders`.
2. Use the following code snippet in `$lib/icons/index.ts`

```ts
const GPIcons = import.meta.glob('/src/lib/icons/google-play/*.png', {
  eager: true,
  import: 'default',
  query: '?url'
}) as Record<string, string>;

export function getGPFallbackIcon(type: ApplicationType) {
  return (
    GPIcons[
      `/src/lib/icons/google-play/${typeof type === 'string' ? type : ApplicationType[type]}.png`
    ] ?? ''
  );
}
```

Note: the `query: '?url'` option is probably necessary for what I am looking for, and may not be supported by the version of Vite we are using. If you run into issues with that being unavailable, please disregard this review comment.

src/routes/(unauthenticated)/(google-play)/user-data/[productId]/+layout.server.ts
```diff
@@ -0,0 +1,83 @@
...
+      }
+    }
+  });
+
+  if (!product) throw error(404);
+
+  const developer = product.Store.GooglePlayTitle?.trim() || 'Unknown developer';
+  const longDesc =
+    translatedManifest['full_description.txt'] || translatedManifest['description.txt'] || '';
+  const fallbackIcon = getAppFallbackIcon(product.Project.ApplicationType?.Id);
```
Comment (FyreByrd) [inline src/routes/(unauthenticated)/(google-play)/user-data/[productId]/+layout.server.ts]: Why are you treating this as if ApplicationType can be null? Because it most emphatically never will be (just check the DB schema).

src/routes/(unauthenticated)/(google-play)/user-data/[productId]/+layout.server.ts
```diff
@@ -0,0 +1,83 @@
...
+  ]);
+
+  const product = await DatabaseReads.products.findUnique({
+    where: { Id: fetchedManifest.productId },
+    select: {
+      Id: true,
+      Project: {
+        select: {
+          Name: true,
+          ApplicationType: { select: { Id: true } }
```
Comment (FyreByrd) [inline src/routes/(unauthenticated)/(google-play)/user-data/[productId]/+layout.server.ts]: Since all you need is the id, you can simplify this by selecting `TypeId` instead of doing the nested query.

src/lib/products/server.ts
FyreByrd, line 1: A few changes I would recommend adding. You will need to update some code in other places as well to accommodate these changes.

```ts
/**
 * normalize an RGB hex color string to a length of 6+ with no leading #
 * eg. fb9 => f0b090
 */
function normalizeColorString(color: string = '#1c3258' /* default to site primary theme color */) {
  const c = color.replaceAll('#', '');
  return c.length < 6 ? `${c[0]}0${c[1]}0${c[2]}0` : c.substring(0, 6);
}

// on line 183 in `manifestSchema`
    color: v.pipe(
      v.string(),
      v.transform((s) => {
        const colors = s.trim().split('\n');
        return {
          light: normalizeColorString(colors[0]),
          dark: normalizeColorString(colors[0].at(-1))
        };
      })
    ),

// on line 233 in `getLatestManifest`
  try {
    const iconURL = new URL(manifest.icon, baseUrl);
    // empty manifest icon before fetching.
    manifest.icon = '';

    const iconCheck = await fetch(iconURL, { method: 'HEAD' });

    // if icon is reachable, set manifest icon to the updated url
    if (iconCheck.ok) {
      manifest.icon = iconURL.href;
    }
  } catch {
    /* empty */
  }
```

src/routes/(unauthenticated)/(google-play)/user-data/[productId]/+page.svelte
```diff
@@ -0,0 +1,285 @@
...
+              <input
+                id="email"
+                type="email"
+                placeholder={m.udm_email_placeholder()}
+                class="input h-11 w-full border border-base-300 text-base focus:border-primary focus:outline-primary sm:text-sm"
+                name="email"
+                bind:value={$deleteForm.email}
+              />
+              <p class="text-[10px] text-base-content/70">{m.udm_email_hint()}</p>
+              {#if $deleteErrors.email}
```
Comment (FyreByrd) [inline src/routes/(unauthenticated)/(google-play)/user-data/[productId]/+page.svelte]: This error message is showing up immediately, and does not disappear after I input an email address.

src/routes/(unauthenticated)/(google-play)/user-data/[productId]/+page.svelte
```diff
@@ -0,0 +1,285 @@
...
+            {#if $deleteMessage?.error}
+              <p class="text-error text-xs leading-tight">{$deleteMessage.error}</p>
+            {/if}
+          </form>
+        </div>
+      </div>
+
+      <p class="mt-4 text-center text-xs text-base-content/70">
+        {m.udm_footer_text({ appName: app.name })}
+        <a class="link link-primary no-underline hover:underline" href="/support">
```
Comment (FyreByrd) [inline src/routes/(unauthenticated)/(google-play)/user-data/[productId]/+page.svelte]: This link goes to a non-existent page. We will need to ask Chris what he would want here...

src/routes/(unauthenticated)/(google-play)/user-data/[productId]/confirm/+page.server.ts
```diff
@@ -0,0 +1,229 @@
...
+
+    const normalizedEmail = form.data.email.trim().toLowerCase();
+    const productId = form.data.productId;
+
+    const code = randomInt(100000, 1000000).toString();
+    const expiresAt = new Date(Date.now() + 10 * 60 * 1000); // 10 minutes from now
+    const now = new Date();
+
+    try {
+      await prisma.$transaction(
```
Comment (FyreByrd) [inline src/routes/(unauthenticated)/(google-play)/user-data/[productId]/confirm/+page.server.ts]: I'm not sure what you're trying to accomplish with this transaction. It looks similar to what you have in `+page.server.ts` one level up. Maybe you could extract the similar functionality into a shared function call? Regardless, I think I'm going to need some explanatory comments. You also have a send code handler in both files.

src/routes/(unauthenticated)/(google-play)/user-data/[productId]/confirm/+page.server.ts
```diff
@@ -0,0 +1,229 @@
...
+        `Create data deletion request task for Product #${productId}`,
+        {
+          type: BullMQ.JobType.UserTasks_DeleteRequest,
+          scope: 'Product',
+          productId,
+          requestId: userChange.Id,
+          comment: userChange.Change ?? undefined,
+          operation: {
+            type: BullMQ.UserTasks.OpType.Create,
+            roles: [RoleId.OrgAdmin]
```
Comment (FyreByrd) [inline src/routes/(unauthenticated)/(google-play)/user-data/[productId]/confirm/+page.server.ts]: Sorry for the confusion, but this should be 

```js
{
  type: BullMQ.UserTasks.OpType.Update,
  targetRole: RoleId.AppBuilder
}
```

FyreByrd (changes requested):
I think the end is in sight. I included snippets in my nitpicky comments so you can see what I'm looking for in those cases.

1. The form error messages are showing up on page load, and do not disappear when I give input.
2. The task pages for the data deletion requests should be a separate PR probably. I will ask Chris about this, but based on time constraints with the end of the semester I may just handle that myself after this is merged.
3. The deletion request page has a link to a non-existent `/support` page. We can discuss during the meeting what to do with this.

src/lib/google-play/udm-theme.css
James-Voight: its a visual decision. I used OKLCH for the light tints because mixing toward white looked more perceptually even, but used sRGB for the dark shades because it kept the manifest color and made it darker. with OKLCH it looked more blue instead of green

src/routes/(unauthenticated)/(google-play)/user-data/[productId]/+page.svelte
Comment (James-Voight) [reply src/routes/(unauthenticated)/(google-play)/user-data/[productId]/+page.svelte]: Ah, this was part of the mock up and I forgot to follow up about it so it has just stayed there

src/lib/google-play/udm-theme.css
FyreByrd: Noted. Also, one of my recommended changes would allow you to use a dark-mode specific color provided by the manifest.

src/routes/(unauthenticated)/(google-play)/user-data/[productId]/+page.svelte
Comment (FyreByrd) [reply src/routes/(unauthenticated)/(google-play)/user-data/[productId]/+page.svelte]: Ok. We could probably just remove it then. Or wait until today's meeting and discuss it there.

src/routes/(unauthenticated)/(google-play)/user-data/[productId]/+layout.server.ts
Comment (James-Voight) [reply src/routes/(unauthenticated)/(google-play)/user-data/[productId]/+layout.server.ts]: just checked the schema again and your right, I'll fix it

src/routes/(unauthenticated)/(google-play)/user-data/[productId]/confirm/+page.server.ts
Comment (James-Voight) [reply src/routes/(unauthenticated)/(google-play)/user-data/[productId]/confirm/+page.server.ts]: The transaction is just to expire older pending codes and either update the latest request or create a new one, but it’s duplicated with the parent page. I’ll pull that into a shared helper and add some comments.

src/routes/(unauthenticated)/(google-play)/user-data/[productId]/+page.svelte
Comment (James-Voight) [reply src/routes/(unauthenticated)/(google-play)/user-data/[productId]/+page.svelte]: That should only show after submit. The issue is also there for the Cloudflare validation. I'll change it so that it clears when the email changes. I was kind of waiting for Josiah to update the cloudflare but I'll do it now

FyreByrd:
Closing in favor of #1509

PR #1477, opened 2026-03-11
Add background process to rename transitions

FyreByrd:
#1468 introduced a background process to associate builds and releases with records in ProductTransition.
This process relies on transitions being named correctly, an assumption that is broken in production by the leftovers of S1.

This PR adds a background process similar to that in #1475 that will slowly work on making sure all of the transitions that would impact the association process are renamed to fit.

## Summary by CodeRabbit

* **Improvements**
  * Staged migration execution with clearer, per-step progress and safer orchestration.
  * Real-time progress updates and UI notifications during migrations for better visibility.

* **New Features**
  * Email notification templates added to inform admins when a migration step finishes and can be removed.


chrisvire (approved):
👍

PR #1518, opened 2026-08-20
Add type query string parameter

chrisvire:
In Google Play, there are two separate configuration fields for account deletion and data deletion.  With `/user-data`, it handles both and defaults to deleting data. With this query string parameter, we can fill in the data-safety.csv with separate URLs  for each. It will go to the same page, but default accordingly.

## Summary by CodeRabbit

* **New Features**
  * Added support for selecting the data deletion type through a URL query parameter.
  * Preserved the selected deletion context when navigating between the data management and About pages.
  * Maintained the current language and query parameters across related links.

* **Bug Fixes**
  * Improved initialization of the deletion form when a valid deletion type is provided.

### DIFF
--- a/src/routes/(unauthenticated)/(google-play)/user-data/[productId=uuid]/+page.server.ts
+++ b/src/routes/(unauthenticated)/(google-play)/user-data/[productId=uuid]/+page.server.ts
@@ -6,7 +6,7 @@ import { valibot } from 'sveltekit-superforms/adapters';
 import * as v from 'valibot';
 import type { Actions, PageServerLoad } from './$types';
 import { env } from '$env/dynamic/private';
-import { deletionTypes, localizedEmailSchema, type DeletionType } from '$lib/google-play';
+import { type DeletionType, deletionTypes, localizedEmailSchema } from '$lib/google-play';
 import { m } from '$lib/google-play/paraglide/messages';
 import type { Locale } from '$lib/google-play/paraglide/runtime';
 import { saveDeleteRequestVerificationCode } from '$lib/google-play/server';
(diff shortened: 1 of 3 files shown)


FyreByrd (changes requested):
Works as expected. Just need to fix the lint issue pointed out by CodeRabbit. Not sure why it isn't showing up in CI... it is for sure showing up locally when I open the file.

To consider:
- Navigating back from the confirmation page doesn't preserve user choice in search params, probably not necessary though...
- Why not `history.back` instead of a complicated goto? (I know the goto is how it was originally...)
- If the user chooses a different option for some reason, goes to the about page, then goes back to the main page, their new choice is not preserved. I don't know how likely this is to happen in the wild though...

chrisvire:
> Works as expected. Just need to fix the lint issue pointed out by CodeRabbit. Not sure why it isn't showing up in CI... it is for sure showing up locally when I open the file.

It isn't showing up as a warning for me when opening the file.  Maybe I have it setup incorrectly??

It did show up in the GHA, but as a warning so it still passed.

> 
> To consider:
> 
> * Navigating back from the confirmation page doesn't preserve user choice in search params, probably not necessary though...
> * Why not `history.back` instead of a complicated goto? (I know the goto is how it was originally...)

I considered that option. Didn't want to depend on browser history. What if for some reason they when directly the the about page??

> * If the user chooses a different option for some reason, goes to the about page, then goes back to the main page, their new choice is not preserved. I don't know how likely this is to happen in the wild though...

Yeah, that is more likely that landing directly on the about page. But they would still lose their choice with using history.back?  

All corner cases which are not all that destructive. They just update their choice. I tried to handle the most common case (click about and click back without first making a choice).

PR #1526, opened 2026-08-26
Improve localization in UDM pages

chrisvire:
* have "show more" / "show less" in language selector which allow users who have installed the app from other Google Play language stores (other than the Listing languages provided with the app) to select their language
* localize the turnstile widget (map what can be and fallback to English)
* try to disable the automatic browser offering to translate the page and allow the user to select localization (browser translation can still be done ... except for turnstile which is in an iframe)

## Language selector with more and less

image

image

## Localization of turnstile

image

## Summary by CodeRabbit

* **New Features**
  * Locale selectors expand and collapse additional languages with localized controls and clearer duplicate names.
  * Turnstile widgets display in the selected language when supported.
  * Added visual indicators for expandable language lists.
* **Bug Fixes**
  * Prevented unnecessary locale redirects and improved unsupported-locale handling.
  * Disabled unavailable “About app” actions and empty-email submissions.
* **Localization**
  * Added “Show less” translations across supported languages.


src/lib/google-play/turnstile.js
```diff
@@ -1,14 +1,81 @@
...
+// https://developers.cloudflare.com/turnstile/reference/supported-languages/
+const LANGUAGE_OVERRIDES = /** @type {Record<string, string>} */ ({
+  'iw-IL': 'he', // Hebrew: Google Play uses the legacy ISO 639-1 code
+  'no-NO': 'nb', // Norwegian: Turnstile only supports Bokmål
+  'fil': 'tl-ph', // Filipino: Ternstile only supports Tagalog (better than defaulting to English)
+  'zh-TW': 'zh-tw', // Chinese (Traditional, Taiwan): base 'zh' would default to Simplified
+  'zh-HK': 'zh-tw' // Chinese (Traditional, Hong Kong): Turnstile has no zh-hk, zh-tw is closer than Simplified
+});
+
+const TURNSTILE_BASE_LANGUAGES = new Set([
```
FyreByrd, line 11: Slight nitpick, but it would be helpful for these to be ordered...

src/lib/google-play/turnstile.js
FyreByrd, line 1: There are still some languages in the locale files that are not supported here, best as I can tell:
af, am, az, be, bn, ca, et, eu, gl, gu, hr, hy, id, is, ka, kk, km, kn, ky, lo, lb, mk, ml, mn, mr, my, ne, nl, pa, rm, si, sq, sw, ta, te, ur, zu

Not sure what level of support we need for these...

chrisvire:
FYI, the screenshot of SAB in the doc is from changes that will be coming in 14.5

image

FyreByrd (changes requested):
Show less could be used on the about page now that it exists (old comment from CodeRabbit in original UDM PR).

I tried a project that had English (en-GB) as one of the provided locales. Since my locale is en-US, I didn't get an about page, even though one ostensibly exists in English, just not American English. There is no readily apparent way to switch to a locale that actually has an about page, though it is possible to do so by hitting show less to reveal only the fully supported locales. I suppose it's fine? But it didn't seem like a phenomenal user experience to me...

src/lib/google-play/turnstile.js
FyreByrd: 37 languages that I assume will be defaulting to English?

chrisvire: Claude reports that there are 34:

af, am, hy-AM, az-AZ, eu-ES, be, bn-BD, my-MM, ca, et, gl-ES, ka-GE, is-IS, kn-IN, km-KH, ky-KG, lo-LA, lv, mk-MK, ml-IN, mr-IN, mn-MN, ne-NP, rm, si-LK, sw, ta-IN, te-IN, sq, gu, kk, pa, ur, zu

chrisvire:
> Show less could be used on the about page now that it exists (old comment from CodeRabbit in original UDM PR).
> 
> I tried a project that had English (en-GB) as one of the provided locales. Since my locale is en-US, I didn't get an about page, even though one ostensibly exists in English, just not American English. There is no readily apparent way to switch to a locale that actually has an about page, though it is possible to do so by hitting show less to reveal only the fully supported locales. I suppose it's fine? But it didn't seem like a phenomenal user experience to me...

I added matching on base variant.

PR #1538, opened 2026-09-15
Track BUILD_MANAGE_VERSION_CODE in Projects.Properties

FyreByrd:
If a project has `BUILD_MANAGE_VERSION_CODE` set to `0` in the publishing properties, then automatic rebuilds will fail for that project; therefore, we needed a way to filter on this.

Changes:
- add `Properties` field to the `Projects` model in the database
- backfill the new field from `publishing_properties` artifact in most recent build
- add API route to update the new field whenever a project is uploaded (pending changes in SAB)
- show field in projects page for super admins only
- disable auto rebuilds toggle if ineligible for automatic rebuilds
  - including explanatory tooltip
- use field to filter rebuildable products in `/software-updates`
- show rebuild setting state in `/projects`
- Miscellaneous fixes:
  - I was running into issues with `upsertJobScheduler` and HMR. I ended up creating a helper function to wrap code in a try catch block with OTEL logging. I am open to splitting it out into a separate PR (and possibly expanding its use...)
  - Per earlier conversation, updated default of `BUILD_MANAGE_VERSION_NAME` to `0`

Screenshot 2026-09-15 at 4 18 09 PM
Screenshot 2026-09-15 at 4 18 22 PM
Screenshot 2026-09-15 at 4 18 54 PM

## Summary by CodeRabbit

- **New Features**
  - Added a publishing-properties API for updating project configuration.
  - Super admins can view project publishing properties on the project details page.
  - Project cards now show whether automatic rebuilding on software updates is enabled.
  - Added guidance explaining the version-management requirement for automatic rebuilding.

- **Bug Fixes**
  - Projects with manually managed version codes are excluded from automatic rebuilds.
  - Automatic rebuilding is disabled and locked when manual version management is detected.
  - Existing project publishing properties are preserved during migration.


src/lib/prisma/schema.prisma
```diff
@@ -282,6 +282,7 @@ model Projects {
   RepositoryUrl           String?
   DateActive              DateTime?        @db.Timestamp(6)
   ImportId                Int?
+  Properties              String?
```
chrisvire, line 285: Just properties?

chrisvire: This is fine since Product.Properties.

FyreByrd: This is consistent with the nomenclature elsewhere in the database (i.e. `Products`, `ProductDefinitions`, etc.)

chrisvire:
> Questions:
> 
> * When auto rebuilds toggle is disabled, should there be an explanatory message?

Yes, we should tell them that BUILD_MANAGE_VERSION_CODE=1 is required for the Software Updates feature

> * What if `BUILD_MANAGE_VERSION_CODE` is set to `0` in a project, but `1` for an individual product in that project? Should the product be eligible for automatic rebuild?

It is not likely to be set there. If product properties have precedence (I can't remember), then yes it shouldn't be eligible.

> * Should there be a way to show that projects are ineligible/don't have the setting enabled in the projects list? (improved UX in finding projects whose settings should be changed)

Yeah, it would be nice to know ahead of time.

